PR #7, opened 2015-03-03
Renamed .get to .check (closes #5)

### LINKED ISSUE #5: Is `Permissions.get` the right name?
`Permissions.get('geolocation')` looks to me like it's going to request the geolocation permission—that is, request access to the user's location.

Maybe `Permissions.check` or `Permissions.has` would be more self-descriptive?

marcoscaceres:


mounirlamouri:
I remember us deciding to move from .get() to .query(). I was actually planning to do that change today. I will go ahead and switch to .query() so we have something better than .get()) for the TAG review.

I am open to discuss switching from .query() to .check(). I don't have strong opinions.

marcoscaceres:
query sounds like you are going to feed it a query, like with querySelector() or an SQL query.

marcoscaceres:
... maybe you are, in a way, doing a query... argh... bike-shedding :(

PR #18, opened 2015-03-25
Have an instance of Permissions on Navigator and WorkerNavigator.

mounirlamouri:
Closes issue 11.

@marcoscaceres PTAL?

### DIFF
--- a/index.html
+++ b/index.html
@@ -153,9 +153,9 @@
         handle permissions on the Web platform. Web APIs have different ways to
         deal with permissions. The [[notifications]] API allows developers to
         request a permission and check the permission status explicitly. Others
-        might only expose the status to web pages. Some, like [[geolocation-API]]
-        will keep the page unaware of the permission associated with the
-        feature.
+        might only expose the status to web pages. Some, like
+        [[geolocation-API]] will keep the page unaware of the permission
+        associated with the feature.
       </p>
       <p>
         Being able to know whether an API call is going to prompt is mandatory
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -412,12 +412,31 @@
     </section>
     <section>
       <h2>
+        Navigator and WorkerNavigator extension
+      </h2>
+      <p>
+        A <a>Permissions</a> instance is exposed on the <code>navigator</code>
+        object for <code>Window</code> and <code>Worker</code> contexts.
+      </p>
+      <dl class='idl' title='[Exposed=(Window)] partial interface Navigator'>
```
marcoscaceres: sorry, why do you have two definitions for Exposed? Don't you just want one, with `[Expose=(Window,Worker)]`?

index.html
```diff
@@ -412,12 +412,31 @@
...
+        <dt>
+          readonly attribute Permissions permissions;
+        </dt>
+      </dl>
+      <dl class='idl' title=
```
marcoscaceres: as above, do you need this redefinition just for the Worker scope?

marcoscaceres: oh, it's WorkerNavigator! d'uh.

index.html
```diff
@@ -445,8 +459,8 @@
         If a developer wants to check multiple permissions at once, the editors
         recommend them to use Promises.all(). It should yield to the same
         result and allow this API to stay simple. If it happens to be a very
-        common use case, it should be easy to extend Permissions.query() to
-        accept a sequence&lt;&gt; too.
+        common use case, it should be easy to extend
+        <code>permissions.query()</code> to accept a sequence&lt;&gt; too.
```
marcoscaceres: I don't think we should set this expectation. I think Promise.all is a great solution to this.

marcoscaceres: In fact, y u no add example of how awesome it is in action?

marcoscaceres:
LGTM.

index.html
mounirlamouri: I am going to open an issue for that.

PR #32, opened 2015-04-27
Boilerplate mail list should be public-webappsec

AFBarstow:
The Mailing List in the boilerplate says p-webapps and since WebAppSec is now publishing this spec, the list should now say p-webappsec.


marcoscaceres:
Thanks!

PR #57, opened 2016-01-19
Mediacapture permissions

alvestrand:
This is my first attempt at incorporating mediacapture-permissions.

It has not been tidy'ed yet, since tidy did other changes to the base spec (it would be better if tidying the base spec was a separate PR).

### DIFF
--- a/index.html
+++ b/index.html
@@ -230,6 +230,18 @@
         <dt>
           midi
         </dt>
+        <dt>
+          camera
+        </dt>
+        <dt>
+          microphone
+        </dt>
+        <dt>
+          speaker
+        </dt>
+        <dt>
+          device-info
+        </dt>
       </dl>
       <p>
         The <code>PermissionName</code> enum defines the list of known
@@ -291,6 +303,63 @@
           </dt>
         </dl>
       </section>
+      <section>
+        <h2>
+          Media Devices
+        </h2>
+        <p>
+          The <code><dfn id=
+          'idl-def-PermissionName.camera'>camera</dfn></code>, <code><dfn id=
+          'idl-def-PermissionName.microphone'>microphone</dfn></code> , and
+          <code><dfn id='idl-def-PermissionName.speaker'>speaker</dfn></code>
+          permissions are associated with permission to use media devices as
+          specified in [[GETUSERMEDIA]] and [[audio-output]].
+        </p>
+        <p>
+          These three permissions have an associated PermissionDescriptor:
+        </p>
+        <dl class='idl' title=
+        'dictionary DevicePermissionDescriptor : PermissionDescriptor'>
+          <dt>
+            DOMString deviceId
+          </dt>
+        </dl>
+        <p>
+          A permission covers access to the device given in the associated
+          descriptor.
+        </p>
+        <p>
+          If the descriptor does not have a deviceId, its semantic is that it
+          queries for access to any device of that class.
+        </p>
+        <p>
+          If the deviceId is the special string "*", it queries for access to
+          all devices of that class, whether they exist at the moment or are
+          attached at a later time. When such a permission has been granted,
+          queries for permission to any specific device MUST return "granted"
+          as long as the deviceId is the ID of a connected device of the given
+          class.
+        </p>
+        <p>
+          The <code><dfn id=
+          'idl-def-PermissionName.device-info'>device-info</dfn></code>
+          permission controls access to names and capabilities of input and
+          output devices.
+        </p>
+        <p>
+          The permission storage identifier for permissions in this section
+          MUST include the top level context's origin and the embedding status.
+        </p>
+        <p>
+          A successful call to the <code>getUserMedia</code> function of
+          [[GETUSERMEDIA]] MUST cause permission to be granted for the returned
+          devices, and MAY cause other permissions to be granted.
+        </p>
+        <p>
+          Stopping a MediaStreamTrack MAY cause permission to be revoked for
+          the associated device.
+        </p>
+      </section>
     </section>
     <section>
       <h2>
@@ -312,7 +381,8 @@
       <ol>
         <li>If the <a>user agent</a> has a <a>permission storage entry</a>
         associated with the <a>permission storage identifier</a> in its
-        permission store, it MUST return the <a>permission storage entry</a>.</li>
+        permission store, it MUST return the <a>permission storage entry</a>.
+        </li>
         <li>Otherwise, it MUST return <code>undefined</code>.</li>
       </ol>
       <p>
@@ -366,7 +436,6 @@
         be asking the user's permission if the caller tries to access the
         feature. The user might grant, deny or dismiss the request.
       </p>
-
       <p>
         The steps to <dfn>retrieve the permission state</dfn> of an <a>origin</a>
         and <a>global object</a> for a given <code>permission</code> are as
@@ -527,9 +596,8 @@
         method is invoked, the <a>user agent</a> MUST run the following steps:
       </p>
       <ol>
-        <il>TODO</il>
+        <li>TODO</li>
       </ol>
-
       <p>
         When the <dfn for='Permissions' title='revoke'><code>revoke()</code></dfn>
         method is invoked, the <a>user agent</a> MUST run the following steps:
@@ -618,6 +686,23 @@
     console.log('Geolocation permission state is ' + result[0].state);
     console.log('Notifications permission state is ' + result[1].state);
   });
+&lt;/script&gt;
+</pre>
+      <p>
+        This example is checking whether the page has the <code>camera</code>
+        permission for the first device in the list. Error checking omitted.
+      </p>
+      <pre class='example highlight'>
+&lt;script&gt;
+  navigator.mediaDevices.enumerateDevices()
+  .then(function(devices) {
+     return navigator.permissions.query({name:'camera',
+                                         deviceId: devices[0].deviceId})
+  })
+  .then(function(result) {
+     console.log('Camera permission to first camera is ' + result.state);
+  }
+  .catch(err =&gt; console.log('Bad things happened: ' + err.name));
 &lt;/script&gt;
 </pre>
     </section>


index.html
```diff
@@ -230,6 +230,27 @@
...
+	  speaker
+	</dt>
+	<dt>
+	  all-cameras
+	</dt>
+	<dt>
+	  all-microphones
+	</dt>
+	<dt>
+	  all-speakers
```
mounirlamouri: I wonder if we could have the "all" cases expressed in the `PermissionDescriptor` instead.

index.html
```diff
@@ -291,6 +312,59 @@
...
+	  </code>
+	  permissions are associated with permission to use media
+	  devices as specified in [[mediacapture-main]].
+	</p>
+	<p>These three permissions have an associated PermissionDescriptor:
+	</p>
+	<dl class='idl' title='dictionary DevicePermissionDescriptor :
+	    PermissionDescriptor'>
+	  <dt>
+	    DOMString deviceId
```
mounirlamouri: How do you express multiple device being available/requested?

index.html
alvestrand: Very possible. The descriptor is only visible as a query parameter, so omitting the deviceId should be OK. Wouldn't even change the IDL, since dictionary elements are by default optional.

index.html
alvestrand: Calling query multiple times and combining them with a permissions.AllOf()?

index.html
```diff
@@ -291,6 +312,59 @@
...
+	  permission controls access to names and capabilities of input
+	  and output devices.
+	</p>
+	<p>
+	  The permission storage identifier for permissions in this
+	  section MUST include the embedder's origin and the embedding
+	  status.
+	</p>
+	<p>
+	  A successful call to the <code>GetUserMedia</code> function of
```
stefhak: lowercase "G"?
Strictly speaking I think there are several ways to call (invoke?) getUserMedia, but perhaps we do not need to be more specific that this.

index.html
```diff
@@ -291,6 +312,59 @@
...
+	  <code><dfn id='idl-def-PermissionName.microphone'>microphone</dfn>
+	  </code>,
+	  and <code><dfn id='idl-def-PermissionName.speaker'>speaker</dfn>
+	  </code>
+	  permissions are associated with permission to use media
+	  devices as specified in [[mediacapture-main]].
+	</p>
+	<p>These three permissions have an associated PermissionDescriptor:
+	</p>
+	<dl class='idl' title='dictionary DevicePermissionDescriptor :
```
stefhak: Should this be _Media_DevicePermissionDescriptor (or do we believe all future devices can use this model)?

index.html
```diff
@@ -291,6 +312,59 @@
...
+	  Media Devices
+	</h2>
+	<p>
+	  The <code><dfn id='idl-def-PermissionName.camera'>camera</dfn></code>,
+	  <code><dfn id='idl-def-PermissionName.microphone'>microphone</dfn>
+	  </code>,
+	  and <code><dfn id='idl-def-PermissionName.speaker'>speaker</dfn>
+	  </code>
+	  permissions are associated with permission to use media
+	  devices as specified in [[mediacapture-main]].
```
stefhak: Aren't permissions for speakers described in mediacapture-output?

stefhak:
I think a media device example should be added. Otherwise, apart from the comments above, this LGTM.

index.html
```diff
@@ -291,6 +312,59 @@
...
+	  permissions have been granted, queries for permission to any
+	  specific device MUST return "granted" as long as the
+	  deviceId is a legal deviceId.
+	<p>
+	  The <code><dfn id='idl-def-PermissionName.device-info'>device-info</dfn></code>
+	  permission controls access to names and capabilities of input
+	  and output devices.
+	</p>
+	<p>
+	  The permission storage identifier for permissions in this
```
stefhak: Trying to parse this. In https://github.com/w3c/mediacapture-main/issues/267 it seems we're zooming in to keying by origin and top-level origin.
Here we we say "origin and embedder's origin". Assuming we have more than one level of iFrame's, this would mean a difference between the two(?).

alvestrand:
PTAL

index.html
mounirlamouri: +1, top level sounds better.

index.html
```diff
@@ -291,6 +303,63 @@
...
+	  If the descriptor does not have a deviceId, its semantic is that
+	  it queries for access to any device of that class.
+	</p>
+	<p>
+	  If the deviceId is the special string "*", it queries for
+	  access to all devices of that class, whether they exist
+	  at the moment or are attached at a later time. When such a
+	  permission has been granted, queries for permission to any
+	  specific device MUST return "granted" as long as the
+	  deviceId is the ID of a connected device of the given class.
```
mounirlamouri: You don't explain what to do in the case of `deviceId` being something else. Is that on purpose. I particularly wonder what we should do if it is the empty string.

index.html
```diff
@@ -291,6 +303,63 @@
...
+	  status.
+	</p>
+	<p>
+	  A successful call to the <code>getUserMedia</code> function of
+	  [[GETUSERMEDIA]] MUST cause permission to be granted
+	  for the returned devices, and MAY cause other permissions to
+	  be granted.
+	</p>
+	<p>
+	  Closing a MediaStreamTrack MAY cause permission to be
```
adam-be: s/Closing/Stopping/ may be a better match to the language in the Media Capture and Streams spec.

index.html
alvestrand: Hm... for the case where it's not '*' and it's not a legal deviceId?

No, that's not specified - I thought the API would have a general text saying "if the parameters are illegal, the result is an error" - but there doesn't seem to be any such language; when retrieving a permission object that doesn't exist, the result is implementation-defined, but always returning "prompt" is one suggested resolution.

I'd suggest an error for illegal descriptors. WDYT?

index.html
```diff
@@ -291,6 +303,63 @@
...
+	associated descriptor.
+	</p>
+	<p>
+	  If the descriptor does not have a deviceId, its semantic is that
+	  it queries for access to any device of that class.
+	</p>
+	<p>
+	  If the deviceId is the special string "*", it queries for
+	  access to all devices of that class, whether they exist
+	  at the moment or are attached at a later time. When such a
```
martinthomson: Is the difference between an absent deviceId and a deviceId of "*" that absence is "devices of this class that are here now" and "*" means "devices of this class, no matter their current presence"

alvestrand: The first one means that getusermedia(...) (no deviceid constraint) will succeed, but you don't know which device you're getting - there might be other cameras you don't have permission to access.

The second means that getusermedia({video: {deviceId: xxx}}) will succeed as long as xxx is a legal deviceid.

(it can fail for other reasons than permission, of course. I'm simplifying).

But as I said, I'm not sure the extra complexity is really worth it.

martinthomson: I think that this way of expressing the "any" vs. "all" distinction is as clear and transparent as mud.

I'm still uncertain about the benefits, but if we do this, we need a more explicit way of making the distinction between specific device, any device and all devices.

index.html
```diff
@@ -291,6 +303,63 @@
...
+	  deviceId is the ID of a connected device of the given class.
+	<p>
+	  The <code><dfn id='idl-def-PermissionName.device-info'>device-info</dfn></code>
+	  permission controls access to names and capabilities of input
+	  and output devices.
+	</p>
+	<p>
+	  The permission storage identifier for permissions in this
+	  section MUST include the top level context's origin and the embedding
+	  status.
```
stefhak: (just to make sure:) Is there a definition of "top level context's origin" somewhere, or is it something that can't be misunderstood?
In the spec there is also a sentence in section 7 that ends "...MAY contain other information like the embedding status or the embedder's origin." Should we add top level context's origin here?

alvestrand: I think we're talking about https://www.w3.org/html/wg/drafts/html/master/browsers.html#top-level-browsing-context - every time I try to follow this chain of references, my head hurts, but I can't quite see how it could be anything else.

index.html
alvestrand: Please suggest text... for instance, we could use the strings "some device" for <empty> and "all devices" for "*" - this revisits the question of what values of a device ID can be used safely for something other than a device id - or we could add another attribute.
My first proposal (separate "all-" permissions) was a bit easier to explain, but did not support the "any" case.

index.html
joeberkovitz: @alvestrand Are you making an attempt to model permission to obtain device labels and metadata via device enumeration, so that the user can make meaningful choices of device?

index.html
mounirlamouri: "all" by default concerns me because it could create issues for browsers that don't do all by default with getUserMedia() calls. However, with permissions.request() we could make this behaviour clearer.

martinthomson@, if you are fine with the possible compat issues that Firefox could run into, I'm fine with removed the distinction between "any" and "all".

alvestrand:
Two changes after discussions:
- Removed the "any" specifier. It was too confusing to reason about what it really meant. The "all" specifier is reasonably easy to reason about, so I described it a little more.
- Removed the scope specification. This is ongoing discussion, and we don't want to add language that makes resolution of ongoing discussions more difficult.

index.html
```diff
@@ -291,6 +303,60 @@
...
+	    DOMString deviceId
+	  </dt>
+	</dl>
+	<p>A permission covers access to the device given in the
+	associated descriptor.
+	</p>
+	<p>
+	  If the descriptor does not have a deviceId, its semantic is that
+	  it queries for access to all devices of that class. Thus, if a query
+          for the "camera" permission with no deviceId returns "granted", the
```
mounirlamouri: nit: alignment issue.

index.html
mounirlamouri: What about this comment?

mounirlamouri:
lgtm with the two comments resolved. Also, could you rebase the commits to one?

index.html
alvestrand: When people have access to the list of devices via enumerateDevices(), they should be able to use that information together with query() to get information about permission to those devices, yes. @joeberkovitz was that what you were asking, or did I misunderstand your question?

index.html
alvestrand: Done.

alvestrand:
One commit, and tidy done. My version of tidy prefers </li> on its own line, it seems. Apologies for the diff noise.

mounirlamouri:
tidy shouldn't do that. You might be using another version of `tidy-html5` or you didn't used the `tidyconfig.txt` file. Note that I saw that `contributing.md` instructions wasn't matching the file name so it could be that.

I know it's painful but I would appreciate if you could clean up the changes.

mounirlamouri:
Thank you for your patience! \o/

alvestrand:
Arrggghhhh - with all the merging, rebasing, tidying and untidying, the last set of changes (removing the * and the special language for top-level origin) have disappeared from the PR :-( - I'll make a new one.

PR #66, opened 2016-03-10
Let each permission refine its algorithms and store data.

jyasskin:
This is now ready to review. See it in action at https://rawgit.com/jyasskin/permissions/allow-choosers/index.html, and a use of the new extension hooks at https://rawgit.com/jyasskin/web-bluetooth-1/permission-api/index.html#permission-api-integration.

### DIFF
--- a/index.html
+++ b/index.html
@@ -294,18 +294,21 @@
         </dt>
         <dd>
           <a>PermissionDescriptor</a> or one of its subtypes.
+          If unspecified, this defaults to <a>PermissionDescriptor</a>.
         </dd>
         <dt>
           A <dfn data-export="">permission storage type</dfn>
         </dt>
         <dd>
           <a>PermissionStorage</a> or one of its subtypes.
+          If unspecified, this defaults to <a>PermissionStorage</a>.
         </dd>
         <dt>
           A <dfn data-export="">permission result type</dfn>
         </dt>
         <dd>
           <a>PermissionStatus</a> or one of its subtypes.
+          If unspecified, this defaults to <a>PermissionStatus</a>.
         </dd>
         <dt>
           A <dfn data-export="">permission query algorithm</dfn>
@@ -318,7 +321,8 @@
           result type</a> instance with the query result. Used by
           <a>Permissions</a>' <a for="Permissions" lt="query">query()</a>
           method and the <a href="#PermissionStatus-update">PermissionStatus
-          update steps</a>.
+          update steps</a>. If unspecified, this defaults to the <a>boolean
+          permission query algorithm</a>.
         </dd>
         <dt>
           A <dfn data-export="">permission request algorithm</dfn>
@@ -333,7 +337,8 @@
           if the request can fail exceptionally. (Merely being denied
           permission is not exceptional.) Used by <a>Permissions</a>' <a for=
           "Permissions" lt="request">request()</a> method, which handles
-          reading and writing the <a>permission store</a>.
+          reading and writing the <a>permission store</a>. If unspecified, this
+          defaults to the <a>boolean permission request algorithm</a>.
         </dd>
         <dt>
           A <dfn data-export="">permission revocation algorithm</dfn>
@@ -342,9 +347,14 @@
           Takes no arguments. Updates any other parts of the implementation
           that need to be kept in sync after an entry is removed from the
           permission store. Triggered by <a>Permissions</a>' <a for=
-          "Permissions" lt="revoke">revoke()</a> method.
+          "Permissions" lt="revoke">revoke()</a> method. If unspecified, this
+          defaults to doing nothing.
         </dd>
       </dl>
+      <p>
+        A <dfn>boolean permission</dfn> is a <a>permission</a> with all types
+        and algorithms defaulted.
+      </p>
       <section>
         <h2>
           Geolocation
@@ -353,46 +363,8 @@
           The <code><dfn id=
           'idl-def-PermissionName.geolocation'>geolocation</dfn></code>
           permission is the permission associated with the usage of the
-          [[geolocation-API]].
+          [[geolocation-API]]. It is a <a>boolean permission</a>.
         </p>
-        <dl>
-          <dt>
-            <a>associated PermissionDescriptor</a>
-          </dt>
-          <dd>
-            <a>PermissionDescriptor</a>
-          </dd>
-          <dt>
-            <a>permission storage type</a>
-          </dt>
-          <dd>
-            <a>PermissionStorage</a>
-          </dd>
-          <dt>
-            <a>permission result type</a>
-          </dt>
-          <dd>
-            <a>PermissionStatus</a>
-          </dd>
-          <dt>
-            <a>permission query algorithm</a>
-          </dt>
-          <dd>
-            Runs the <a>binary permission query algorithm</a>.
-          </dd>
-          <dt>
-            <a>permission request algorithm</a>
-          </dt>
-          <dd>
-            Runs the <a>binary permission request algorithm</a>.
-          </dd>
-          <dt>
-            <a>permission revocation algorithm</a>
-          </dt>
-          <dd>
-            Does nothing.
-          </dd>
-        </dl>
       </section>
       <section>
         <h2>
@@ -402,46 +374,8 @@
           The <code><dfn id=
           'idl-def-PermissionName.notifications'>notifications</dfn></code>
           permission is the permission associated with the usage of the
-          [[notifications]] API.
+          [[notifications]] API. It is a <a>boolean permission</a>.
         </p>
-        <dl>
-          <dt>
-            <a>associated PermissionDescriptor</a>
-          </dt>
-          <dd>
-            <a>PermissionDescriptor</a>
-          </dd>
-          <dt>
-            <a>permission storage type</a>
-          </dt>
-          <dd>
-            <a>PermissionStorage</a>
-          </dd>
-          <dt>
-            <a>permission result type</a>
-          </dt>
-          <dd>
-            <a>PermissionStatus</a>
-          </dd>
-          <dt>
-            <a>permission query algorithm</a>
-          </dt>
-          <dd>
-            Runs the <a>binary permission query algorithm</a>.
-          </dd>
-          <dt>
-            <a>permission request algorithm</a>
-          </dt>
-          <dd>
-            Runs the <a>binary permission request algorithm</a>.
-          </dd>
-          <dt>
-            <a>permission revocation algorithm</a>
-          </dt>
-          <dd>
-            Does nothing.
-          </dd>
-        </dl>
       </section>
       <section>
         <h2>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -244,9 +269,71 @@
         </dt>
       </dl>
       <p>
-        The <code>PermissionName</code> enum defines the list of known
-        permission names.
+        Each enumerator-value in the <a>PermissionName</a> enum identifies a
```
domenic: enumerator-value is a weird term I've never heard before. Is it a C++ thing? Prefer "enumeration value" or "enum value".

index.html
```diff
@@ -244,9 +269,71 @@
...
+        <dfn data-export="">permission</dfn>, which consists of the following
```
domenic: Maybe it's against prevailing style in the file, but `<dfn data-export>` is equivalent and shorter.

index.html
```diff
@@ -244,9 +269,71 @@
...
+        algorithms and types:
       </p>
+      <dl>
+        <dt>
+          An <dfn data-export="">associated PermissionDescriptor</dfn>
```
domenic: Seems like things would be more consistent if this were "A permission descriptor type" to match the next two.

index.html
```diff
@@ -244,9 +269,71 @@
...
+        </dt>
+        <dd>
+          Takes an <a>environment settings object</a> argument and returns a
+          <a>permission storage identifier</a>.
+        </dd>
+        <dt>
+          A <dfn data-export="">permission query algorithm</dfn>
+        </dt>
+        <dd>
+          Takes an <a>associated PermissionDescriptor</a>, an instance of the
```
domenic: I think this would be "an instance of the associated permission descriptor type for this permission". The main thing is instance; right now associated PermissionDescriptor/associated permission descriptor type is a type, not an object.

index.html
jyasskin: It's a funny mix of C++ and WebIDL. Oops. Fixing.

index.html
```diff
@@ -244,9 +269,71 @@
...
+          <a>permission storage type</a> that's currently stored for this
+          <a>permission</a>, and an existing instance of the <a>permission
+          result type</a>, and updates the <a>permission result type</a>
+          instance with the query result. Used by <a for="Permissions" lt=
+          "query">Permissions.query()</a> and the <a href=
```
domenic: Probably also a prevailing style issue but this is pretty bad form; see https://github.com/whatwg/html/issues/826

jyasskin:
I think this is ready for a serious look now, @marcoscaceres @mounirlamouri. Feel free to tell me to split it up into a couple different PRs.

index.html
```diff
@@ -244,9 +269,71 @@
...
+          "#PermissionStatus-update">PermissionStatus update steps</a>.
+        </dd>
+        <dt>
+          A <dfn data-export="">permission request algorithm</dfn>
+        </dt>
+        <dd>
+          Takes the previously-stored <a>permission storage type</a>, an
```
domenic: This paragraph needs a few "instance of"s inserted as well.

index.html
```diff
@@ -244,9 +269,71 @@
...
+          <a>associated PermissionDescriptor</a>, and a newly-created
+          <a>permission result type</a>, shows the user any necessary prompt to
```
domenic: s/shows/and shows

domenic: Or maybe better s/, shows/. Shows

index.html
jyasskin: https://github.com/w3c/permissions/blob/gh-pages/contributing.md says to run tidy, which adds the `=""`. :-/

index.html
```diff
@@ -244,9 +269,71 @@
...
+          try to increase permissions, and updates the <a>permission storage
+          type</a> and <a>permission result type</a> to match. Used by <a for=
+          "Permissions" lt="request">Permissions.request()</a>, which handles
+          reading and writing the <a>permission store</a>.
+        </dd>
+        <dt>
+          A <dfn data-export="">permission revocation algorithm</dfn>
+        </dt>
+        <dd>
+          After <a for="Permissions" lt="revoke">Permissions.revoke()</a>
```
domenic: This paragraph is inconsistent with the other two algorithms. Should be something like "Takes no arguments. Updates any other parts of the implementation that need to be kept in sync after an entry is removed from the permission store. Triggered by Permissions's revoke() method.

index.html
jyasskin: I'd actually lean toward "permission query type", but "associated PermissionDescriptor" was already defined in the file, and that cleanup seemed like something I could plausibly leave to a subsequent PR. This one's already huge.

index.html
```diff
@@ -244,9 +269,71 @@
...
+        Each enumerator-value in the <a>PermissionName</a> enum identifies a
+        <dfn data-export="">permission</dfn>, which consists of the following
+        algorithms and types:
       </p>
+      <dl>
+        <dt>
+          An <dfn data-export="">associated PermissionDescriptor</dfn>
+        </dt>
+        <dd>
+          A subtype of <a>PermissionDescriptor</a>.
```
domenic: Looking at all the examples, it seems like it's often not a subtype. Not sure how best to phrase it; maybe just a parenthatical (or PermissionDescriptor itself if a custom type is not needed).

index.html
jyasskin: Done.

index.html
jyasskin: Done.

index.html
jyasskin: Done.

index.html
jyasskin: Hm, I didn't check the examples before writing this, but I see just:
```javascript
{name:'geolocation'}
{name:'notifications'}
{name:'camera', deviceId: devices[0].deviceId}
```
which are all instances of subtypes of  `PermissionDescriptor`:
```webidl
dictionary PermissionDescriptor {
    required PermissionName name;
};
```

What example are you looking at?

index.html
jyasskin: Done.

index.html
jyasskin: Done.

index.html
domenic: PermissionDescriptor is not a subtype of PermissionDescriptor, at least not in any intuitive sense. I was looking at geolocations and notifications.

jyasskin: Oh, I see. Strictly, types are subtypes of themselves, but not _proper_ subtypes of themselves, but I can word it differently. Maybe `<a>PermissionDescriptor</a> or one of its subtypes`?

domenic: Right, outside of a mathematical context I'd err on the side of assuming subX = proper subX. That sounds pretty good to me.

jyasskin:
See https://rawgit.com/jyasskin/web-bluetooth-1/permission-api/index.html#permission-api-integration for the beginning of Web Bluetooth's use of these new hooks.

raymeskhoury:
Hey Jeffrey, a few random thoughts (some are not solely related to your edits):
-It initially hard to see which parts were external/exposed APIs vs internal constructs. e.g. PermissionDescriptor and PermissionStatus are external JS APIs - they must be complied with precisely whereas PermissionStorage is merely meant to help define storage behavior. Would it be helpful to draw a more clear line between those two groups?
-I also found it a bit hard to draw the distinction between required and recommended behaviors. It seems as though there should be room for UAs to handle permissions in ways that are outside of those currently specified behavior (for example, we have policy settings in Chrome that can override permissions). I don't think this is directly related to your edits.
-In relation to the above two points, perhaps a top-down restructuring of the spec could be helpful? External APIs and high level, required behaviors specified first and per-API behaviors, more subtle recommended behaviors, implementation details, etc. specified after? I'm no spec writer though :)
-I didn't think we were going to draw a distinction between an origin-based and embedding-based permission storage identifier instead leaving it up to the UA. I just pinged Mounir asking about this because we had chatted about it earlier. If it turns out to be the case does the permission storage identifier need to vary per-permission?

jyasskin:
Thanks for the comments!

> It initially hard to see which parts were external/exposed APIs vs internal constructs. e.g. PermissionDescriptor and PermissionStatus are external JS APIs - they must be complied with precisely whereas PermissionStorage is merely meant to help define storage behavior. Would it be helpful to draw a more clear line between those two groups?

Yeah. The traditional way to distinguish these would be to avoid describing `PermissionStorage` in IDL. However, that loses IDL's precision about what's stored, which I think is helping Bluetooth (I should be able to show you that tomorrow). I left a note saying `PermissionStorage` isn't exposed, but if you have other suggestions, I'm happy to take them.

>I also found it a bit hard to draw the distinction between required and recommended behaviors. It seems as though there should be room for UAs to handle permissions in ways that are outside of those currently specified behavior (for example, we have policy settings in Chrome that can override permissions). I don't think this is directly related to your edits.

Right, I think that's pre-existing. In most cases, I think we can say "the UA may customize the permission prompt, including auto-granting and auto-denying it" and "the UA may revoke at any time", and that's enough? I'd like to fix that in a separate PR, to keep this one from growing much more.

> In relation to the above two points, perhaps a top-down restructuring of the spec could be helpful? External APIs and high level, required behaviors specified first and per-API behaviors, more subtle recommended behaviors, implementation details, etc. specified after? I'm no spec writer though :)

Probably. Again, I want to keep rearrangements separate from the API extension.

> I didn't think we were going to draw a distinction between an origin-based and embedding-based permission storage identifier instead leaving it up to the UA. I just pinged Mounir asking about this because we had chatted about it earlier. If it turns out to be the case does the permission storage identifier need to vary per-permission?

Yeah, if it's always `PermissionName` + origin + UA-specific data, I think we can drop the identifier hook. I'll let Mounir confirm and then do that in my next update.

jyasskin:
The identifier hook is gone. PTAL.

mounirlamouri:
At a high level I agree with the changes here. Though, it's a very large CL and I can't realistically review this. Could you split this in a few parts:
- the commit that export the retrieve a permission (CC @tobie)
- the PermissionStorage definition
- the rest of the changes (I might ask you to split more but at least, I will have a better visibility -- keep the commits separated, it might be easier)

Also, could you remove tidy changes? It seems that we have issues with your tidy runs getting other results than mine. You are not the first person. I would like to figure this out before every one changes the spec to match its own tidy runs.

jyasskin:
Ok, I've split out a bunch of separate PRs that build on each other. I haven't yet undone the Tidy changes, and I've squashed the remaining changes in this PR since the individual commits were getting incoherent.

tobie:
> I haven't yet undone the Tidy changes

OK, sorry for bitching in the other PR, then. :/

jyasskin:
I've noticed that [PushMessagingPermissionContext::DecidePermission](https://code.google.com/p/chromium/codesearch/#chromium/src/chrome/browser/push_messaging/push_messaging_permission_context.cc&q=PushMessagingPermissionContext::DecidePermission&sq=package:chromium&type=cs&l=74) checks for `requesting_origin == embedding_origin`, which doesn't fit into the current [permission request algorithm](https://rawgit.com/jyasskin/permissions/allow-choosers/index.html#dfn-permission-request-algorithm) arguments.

If it's ok with you folks, I'd like to get the current PR merged, and then fix up details like this in subsequent smaller patches.

PR #68, opened 2016-03-15
Re-entering agreed fixes to Media permissions

alvestrand:
This PR re-enters the changes that were lost in transition
while entering the addition of the media permissions.


mounirlamouri:
Thanks!

PR #72, opened 2016-03-23
Make Permissions.request() only take a single descriptor.

jyasskin:
This matches the [current Chromium implementation](https://code.google.com/p/chromium/codesearch/#chromium/src/third_party/WebKit/Source/modules/permissions/Permissions.idl), which has a `requestAll()` function to handle a sequence of permissions. A sequence could also be handled with `Promise.all()`, like [`query()` suggests](https://rawgit.com/jyasskin/permissions/no-request-sequence/index.html#h-note1).

### DIFF
--- a/index.html
+++ b/index.html
@@ -99,12 +99,22 @@
       </p>
       <ul>
         <li>
+          <dfn><a href=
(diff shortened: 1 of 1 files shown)


mounirlamouri:
119deb4 lgtm

PR #73, opened 2016-03-23
Define a PermissionStorage type for each permission to store data with.

jyasskin:
https://rawgit.com/jyasskin/permissions/storage-type/index.html#idl-def-PermissionStorage

This is separate from the PermissionStatus that gets returned to users and from
the PermissionDescriptor that users query the permission with.

### DIFF
--- a/index.html
+++ b/index.html
@@ -399,8 +399,7 @@
         <a data-lt='user agent'>User agents</a> MAY use a form of storage to
         keep track of web site permissions. When they do, they MUST have a
         <dfn>permission storage identifier</dfn> which is linked to a
-        <dfn>permission storage entry</dfn>. The <a>permission storage
-        entry</a> MUST be a <a>PermissionState</a> or <code>undefined</code>.
+        <a>PermissionStorage</a> instance or one of its subtypes.
       </p>
       <p>
         To <dfn>get a permission storage identifier</dfn> for a
@@ -420,14 +419,22 @@
           settings</var>' <a>top-level browsing context</a>'s <a>origin</a>
         </li>
       </ol>
+      <dl class='idl' title='dictionary PermissionStorage'>
+        <dt>
+          required PermissionState state
+        </dt>
+      </dl>
+      <p class="note">
+        <a>PermissionStorage</a> instances are never exposed to Javascript.
+      </p>
       <p>
         The steps to <dfn>retrieve a permission storage entry</dfn> of a
         <a>permission storage identifier</a> are as follows:
       </p>
       <ol>
-        <li>If the <a>user agent</a> has a <a>permission storage entry</a>
-        associated with the <a>permission storage identifier</a> in its
-        permission store, it MUST return the <a>permission storage entry</a>.
+        <li>If the <a>user agent</a> has a <a>PermissionStorage</a> associated
+        with the <a>permission storage identifier</a> in its permission store,
+        it MUST return the <a>PermissionStorage</a>.
         </li>
         <li>Otherwise, it MUST return <code>undefined</code>.
         </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -420,14 +419,22 @@
...
+      <dl class='idl' title='dictionary PermissionStorage'>
+        <dt>
+          required PermissionState state
+        </dt>
+      </dl>
+      <p class="note">
+        <a>PermissionStorage</a> instances are never exposed to Javascript.
+      </p>
       <p>
         The steps to <dfn>retrieve a permission storage entry</dfn> of a
```
raymeskhoury: Should this be "PermissionStorage"?

index.html
```diff
@@ -514,43 +523,66 @@
         </dt>
       </dl>
       <p>
-        A <code>PermissionStatus</code> instance has an associated
-        <var>permission</var> that is a <a>PermissionDescriptor</a>.
+        <code>PermissionStatus</code> instances are created with the following
```
raymeskhoury: I'm curious why it's useful to define this as an internal slot as opposed to just saying that it has an associated PermissionDescriptor? It seems a bit more confusing to me.

index.html
raymeskhoury: Should this be "PermissionStorage"? (and below as well)

index.html
```diff
@@ -420,14 +419,22 @@
           settings</var>' <a>top-level browsing context</a>'s <a>origin</a>
         </li>
       </ol>
+      <dl class='idl' title='dictionary PermissionStorage'>
```
raymeskhoury: I wonder if there is a way we can make it even more clear this is an internal-only structure.

jyasskin: I can make it look like:

```
dictionary PermissionStorage {
    // PermissionStorage is just an explanatory device.
    // Instances are never received from or passed to Javascript code.
    required PermissionState state;
};
```

Otherwise, I'm open to suggestions, but I do find that having a concrete list of properties, however expressed, makes it easier to follow specifications.

raymeskhoury: That sounds good. I was going to suggest changing the name as well, but I couldn't think of a way of making it clear in the name.

index.html
jyasskin: `PermissionStatus`es are going to have two internal slots once we get to #66, in case that makes it look a little better: https://rawgit.com/jyasskin/permissions/allow-choosers/index.html#idl-def-PermissionStatus.

I'm finding in [Web Bluetooth](https://webbluetoothcg.github.io/web-bluetooth/#bluetoothservicedatamap) that having a bunch of internal slots makes it easier to understand the purpose of each bit of associated data, and that `foo@[[bar]]` is simpler to type and link than "the bar associated with `foo`". The `PermissionStatus` slots are simple enough that I can eliminate them without needed a lot of extra wording, but I do somewhat prefer them.

@annevk and @domenic may also have opinions about which to use.

index.html
jyasskin: This is the sort of cleanup that I want to do, but in a separate patch, so this series doesn't grow any more. We currently have [retrieve a permission storage entry](https://rawgit.com/jyasskin/permissions/allow-choosers/index.html#dfn-retrieve-a-permission-storage-entry) that's only called from [retrieve the permission storage](https://rawgit.com/jyasskin/permissions/allow-choosers/index.html#dfn-retrieve-the-permission-storage), so maybe I can eliminate an unnecessary layer entirely, instead of just renaming it.

index.html
jyasskin: Done.

index.html
annevk: I don't care strongly at the moment. Hopefully going forward IDL will provide formalism around this so you can define the layout of an object, including its internal slots, upfront.

PR #75, opened 2016-04-09
Convert to Bikeshed

jyasskin:
Compare https://rawgit.com/jyasskin/permissions/bikeshed/index.html with https://w3c.github.io/permissions/ to see if I broke anything. The biggest benefit of this is that we get automatic links to and from other specs.

If someone else merges this, don't squash it completely. The 'move' commit needs to stay separate from the conversion commit so it's possible to follow the history.


mounirlamouri:
/CC @marcoscaceres 

I believe respect does linking in and out. What doesn't work exactly?

jyasskin:
Both linking directions are more difficult than necessary with ReSpec: to link out, you need the [Dependencies](https://w3c.github.io/permissions/#dependencies) section and following a link takes two clicks. Even when Shepherd doesn't include the other spec to make the out-link automatic, the extra links you need are easier to write (https://github.com/w3c/permissions/pull/75/commits/49c5804ac23e02cd0df628cc3d89fadc49926b41#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR25), and folks are adding that kind of thing to Shepherd fairly quickly right now.

To let other specs link in, you have to manually add all the [dfn bits](https://github.com/tabatkins/bikeshed/blob/master/docs/dfn-contract.md) that Shepherd expects in order to let the other specs auto-link, and that can be extra-tricky for IDL things. It'd be possible to teach ReSpec to include this stuff, but Bikeshed's already there.

ReSpec also causes a FOUC that breaks reloads while you're iterating on a section. Try visiting https://w3c.github.io/permissions/#status-of-a-permission and pressing 'reload': you won't be looking at the right section anymore.

ReSpec also doesn't have any significant benefits anymore: the ability to save and then reload without an intermediate build step is provided by `bikeshed watch`.

marcoscaceres:
All of these points are valid. I'm ok with the switch, but please try to keep the document source organized (I.e., tidy'd in some way). BikeShed's use of semantic like breaks makes it really hard to review changes and move stuff around the document.

mounirlamouri:
Thanks for the explanations Jeffrey :)

TBH, I'm not a big fan of switching to bikeshed because ultimately, it sounds like a vim VS emacs discussion but given that you have everything lined up and Marcos is fine with it, I'm not going to block you so fool free to merge.

PR #79, opened 2016-04-14
Remove the incorrect paragraph about request() and revoke() having been removed from the spec.

jyasskin:
This paragraph keeps confusing people, since these functions are in fact specified. Removing the paragraph doesn't guarantee we'll keep the functions: it just makes the spec self-consistent.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -85,13 +85,6 @@ spec: webidl
     this document, please contact the editors or file an issue. We would
     love to hear about it.
   </p>
-  <p>
-    The initial intent of this document was to allow web applications to
-    request and revoke permissions explicitly in addition to querying the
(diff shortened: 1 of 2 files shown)


jan-ivar:
Is the controversy over?

jan-ivar:
Consistency could equally be reached by removing the `request` feature.

jyasskin:
Controversy isn't an excuse for having a self-inconsistent specification. Removing non-normative text is the quickest way to consistency, and even if a subsequent discussion concludes that we should remove both request() and revoke() too, I don't think we'll wind up re-adding this paragraph.

marcoscaceres:
LGTM

PR #80, opened 2016-04-14
Export dfn-type definitions that other specs should be able to use.

jyasskin:
* permission store
* permission storage identifier
* retrieve the permission storage
* boolean permission
* boolean permission query algorithm
* boolean permission request algorithm

Web Bluetooth is about to use the first 3 of these. It also uses a couple other terms, but I think it's a bug that it needs to, so I don't want to export them.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -212,7 +212,7 @@ spec: webidl
     </dd>
   </dl>
   <p>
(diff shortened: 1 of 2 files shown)


marcoscaceres:
r+

PR #81, opened 2016-04-17
Rename the "associated PermissionDescriptor" to the "permission descriptor type".

jyasskin:
As @domenic requested in #66. See https://rawgit.com/jyasskin/permissions/rename-associated-permission-descriptor/index.html#permission-descriptor-type.

I wanted to use "permission query type", but the type is also used to describe the permission being requested in `request()`, so descriptor fits.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -112,11 +112,11 @@ spec: webidl
     };
   </pre>
   <p>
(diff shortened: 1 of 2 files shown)


mounirlamouri:
lgtm, thanks! :)

PR #82, opened 2016-04-17
Rename PermissionStatus to PermissionResult.

jyasskin:
Preview at https://rawgit.com/jyasskin/permissions/rename-permission-status/index.html#permissionresult.

The idea is that the result of `query()` isn't just the status of a permission; it can include extra data like, for Bluetooth, the list of allowed devices that match the query. Do folks think this is a good change? This changes user-visible behavior if folks are modifying `PermissionStatus.prototype`, so I should check that nobody's actually doing that.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -160,8 +160,8 @@ spec: webidl
       A <dfn export>permission result type</dfn>
     </dt>
     <dd>
-      {{PermissionStatus}} or one of its subtypes.
-      If unspecified, this defaults to {{PermissionStatus}}.
+      {{PermissionResult}} or one of its subtypes.
+      If unspecified, this defaults to {{PermissionResult}}.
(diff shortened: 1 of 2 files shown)


mounirlamouri:
This is shipping in two browsers. Furthermore, this was meant to provide the status of a permission that you might get from various places like `query()`, `request()` or `revoke()`. It's not sure `PermissionResult` is the best name for all these cases.

jyasskin:
@slightlyoff convinced me not to push on this. "Status" still isn't a great name: it's confusable with [`PermissionState`](https://w3c.github.io/permissions/#enumdef-permissionstate) and seems to represent the global status of a permission rather than the result of a particular `query()` or `request()` operation, but there are worse warts in the web platform.

Issue #86, opened 2016-04-21
Model temporary permissions better


jyasskin:
Right now, we have a single [permission store](https://w3c.github.io/permissions/#permission-store) that has basically one entry per permission per origin. This isn't rich enough to model UAs that can grant permission during the lifetime of a single tab. So, what model should we have instead?

I believe the basic cross-browser idea is that each [environment settings object](https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object) (tab visiting a particular page) has some grants and denials for various permissions ("local storage"). Each origin (roughly) also has some grants and denials stored in the browser as a whole ("global storage"). It may also be possible in some cases to continue using an object-capability (a `MediaStream` or series of `watchPosition` callbacks) even though no permission is stored even locally.

* When an environment settings object is created, it gets its initial local storage from the global storage for its origin.
* When permissions are requested inside a tab and the user makes a decision, that decision is stored in the settings object's storage and maybe also in the global storage, depending on the UA and the user's precise behavior. (e.g. Chrome generally records decisions in both the local and global storage, while Firefox defaults to recording grants in only the local storage unless the user opts into global storage.)
* When permissions are revoked through UI, Chrome only touches the global storage (this is arguably a bug), while Firefox sometimes revokes both the global and all local storages (location), but sometimes revokes only the global and current local storage (camera).

So the question is how to specify all of this flexibility.

alvestrand:
I don't agree that this is the model .... the permissions store is keyed by origin, not by session. So the current spec is only the global store from your description.
It would be interesting to investigate the concept of a session-keyed store in addition to the origin-keyed store. But I don't think it exists in the spec now.

jyasskin:
Yeah, sorry for the lack of clarity. I'm trying to figure out the model we ought to put in the spec. The model that's currently in the spec is wrong.

jan-ivar:
Unfortunately, the poor terminology used in this spec casts doubt over any consensus achieved so far.

My advice would be to shore up the existing language and terminology first, to salvage consensus, before attempting to add additional complexity to this.

Clear up the existing definitions in the document to reflect that they are *persistent permissions* as written, so that we have a minimum shared basis on what has been agreed to so far.

jyasskin:
I see browsers doing 3 distinct things with a series of calls to `navigator.geolocation.watchPosition()`.

[...]

Issue #74, opened 2016-04-01
Review of jyasskin's draft


annevk:
I reviewed https://rawgit.com/jyasskin/permissions/allow-choosers/index.html
1. Everything that is associated with a permission should default to the "binary" (rename to boolean?) options. There's no need to repeat that for each of them.
2. Since the majority are boolean, there should be a simple way to get and set the state of boolean permissions. See https://notifications.spec.whatwg.org/ for how simple this can be.
3. permissions.request() should not be available in workers I think.

jyasskin:
I've done (1) in #66; the rest should wait for separate PRs.

marcoscaceres:
2 above is duplicate of https://github.com/w3c/permissions/issues/42 ... 1 was done in #66, and 3 doesn't apply as there is no .request().

Issue #33, opened 2015-05-06
Define underlying infrastructure


annevk:
See e.g. Notifications API Standard or Storage Standard.

Specifications will need access to the underlying algorithms as to what the status of a permission is. They might also want to subscribe to change notifications as this might impact existing objects (e.g. for notifications we'd likely remove the notifications).

jyasskin:
Are the [permission request algorithm](https://w3c.github.io/permissions/#permission-request-algorithm) and [permission revocation algorithm](https://w3c.github.io/permissions/#permission-revocation-algorithm) sufficient for the "change notifications", or do you want something more tailored?

Note that the revocation algorithm needs to pass the descriptor down and let the permission itself decide how to update storage. Assume I've done that when answering. :smiley:

annevk:
I think one big problem is that origins are not accounted for. E.g., https://w3c.github.io/permissions/#dom-permissions-query does not state where it gets its origin value.

I think https://w3c.github.io/permissions/#boolean-permission-query-algorithm also needs to take an origin.

How do I store a permission for an origin?

(And note that for notifications I need to be able to do some of this synchronously. Although I suppose I could simply say "wait" or "pause" and reference HTML.)

jyasskin:
#91 should let you do all of this synchronously using exported dfns, and has examples for the wording you'd use.

The origins come somewhat implicitly from [get a permission storage identifier](https://rawgit.com/jyasskin/permissions/update-permission/index.html#get-a-permission-storage-identifier), and we might have to change that up again due to https://noncombatant.github.io/permission-delegation-api/. I'm definitely happy to take suggestions on how to make it all clearer. My best guess at the moment is to find a better name for "permission storage identifier", but I don't have a concrete name in mind.

annevk:
I see. "key" would already be clearer I think. Since this is a rather traditional key-value store if I'm not mistaken.

marcoscaceres:
Closed by https://github.com/w3c/permissions/pull/91

PR #91, opened 2016-04-25
Define an algorithm to update the permission storage.

### LINKED ISSUE #84: Why is "retrieve the permission storage" not in "Permission store"?

jyasskin:
Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/update-permission/index.bs#storage-manipulation.

This contributes to #33, #74, and #86, and fixes #84.

@annevk, how do you like the wording that would wind up in [[notifications]]?


index.bs
```diff
@@ -439,6 +440,157 @@ spec: webidl
...
+
+    <div class="example" id="example-storage-manipulation">
+      <p>
+        The [[notifications]] API can check for the user having granted
+        permission using:
+      </p>
+      <blockquote>
+        If the {{"notifications"}} <a>permission's state</a> is not
+        {{"granted"}}, reject <var>promise</var> with a {{TypeError}} exception
+        and terminate these substeps.
```
annevk: I don't understand how this works without the key being in the mix somewhere. This reads like "notifications" permission's state is global.

jyasskin: In all the uses of permission state, I couldn't find one where it pulled the origin out of anything but the current/entry/incumbent settings object, so I factored that into ["name" permission's storage](https://rawgit.com/jyasskin/permissions/update-permission/index.html#permissions-storage) so we don't have to re-state it everywhere. I could have you pass the current settings object explicitly, if you think that makes the usage clearer.

Given https://noncombatant.github.io/permission-delegation-api/, it's not guaranteed that the key's always going to use the *origin* of the current settings object: it might use the origin of its embedder along with some flags on the iframe instead.

annevk: Wouldn't that require it to be passed in?

annevk: Also, do all legacy APIs always use the current settings object in browsers? Never entry/incumbent/relevant?

jyasskin: The current realm is just ambiently available, I think. Nothing in WebIDL says it goes away when you call another algorithm.

The origin is the same for all three, so until folks start using other aspects of the settings object, entry/incumbent/relevant doesn't matter. We'll have to ask @raymeskhoury and @palmer if they've picked one for checking embedding.

annevk: It doesn't have to be, given `document.domain`. But maybe you always ignore the domain field...

annevk: Note that IDL is notoriously vague in this area. Most of it is in HTML and that's being cleaned up too.

jyasskin: I always forget how awful `document.domain` is. That said, @lgarron pointed out that Chrome does something even weirder with permission storage: a bunch of permissions are stored with a wildcard domain, so multiple origins can match the same entry in the permission store: https://code.google.com/p/chromium/codesearch#search/&q=function:ContentSettingsPattern::FromURL$%20file:.cc

I'm now inclined to put the lookup of the settings object in [get a permission storage identifier/key](https://rawgit.com/jyasskin/permissions/update-permission/index.html#get-a-permission-storage-identifier) and downgrade the MUST to a SHOULD to let the value match current practice. We can tighten it back up as we write tests for current browser behavior and/or fix browsers to do something more sensible.

lgarron: Just to add a bit more, some of the permissions are stored as a combination of `request origin + top-level origin` when they are requested from a frame.

I believe the motivations for wildcards was that you could go into settings and enable a permission for `*.example.com` and never have to worry about individual permissions for subdomains (either for developers of power users).
In addition, if the permission only takes into account the requesting origin (and not the top-level origin if the request comes from a frame), then a permission granted to an origin in one place can be automatically granted in other frames for that origin.
As a concrete example, this would be desirable if you consider fullscreen a permission and want to permit all YouTube video frames to be able to go fullscreen after granting fullscreen once on `youtube.com`.
However, what Chrome does should not be considered a good model. It was constructed ad-hoc, depends on the permission in question, and is in the process of being converted to simply origin-scoped.

annevk: Gecko used to have a domain-based permission model as well, even across HTTP/HTTPS. We recently fixed that to make it origin-based.

(My comment about `document.domain` was slightly misleading. With `document.domain` those settings objects can have different origins, period. They would compare as equal for "same origin-domain", but might not always do so, and I believe we typically use "same origin" for permissions though as you noted there is still some variance there.)

alvestrand:
About the key for the store (request origin vs top level origin vs both):
The WEBRTC WG had a long discussion on this, and concluded that granting access to a request origin independent of top level origin was not viable; it would allow any app that got permission somehow to utilize that permission in all contexts (think "one-pixel tracker gif with a microphone").
So one version of our spec required double keying (any new combination of top-level origin and request origin had to request permission separately).
After that version, we were informed about an initiative to push for granting permissions in general to top-level origins only and require explicit delegation for the permission to be usable by embedded iframes; that's where people's heads seem to be at the moment.
I'd love to have the decision on that be made for the platform, not for each spec on its own.

jyasskin:

Issue #90, opened 2016-04-23
Specify how request and revoke work on push and midi permissions.


jyasskin:
The ["push"](https://w3c.github.io/permissions/#push) and ["midi"](https://w3c.github.io/permissions/#midi) permissions have a simple ordering in their descriptors, which we haven't fleshed out into their storage or their request and revoke algorithms. Specifically, `{name:"push", userVisibleOnly:false}` is strictly stronger (allows more page behaviors) than `{name:"push", userVisibleOnly:true}`, and `{name:"midi", sysex:true}` is strictly stronger than `{name:"midi", sysex:false}`.

So, if the page calls `navigator.permissions.request({name:"push", userVisibleOnly:false})` and the user accepts, should `navigator.permissions.query({name:"push", userVisibleOnly:true})` return `{state:"granted"}`? Probably so.

If `navigator.permissions.query({name:"push", userVisibleOnly:false})` returns `{state:"granted"}` and the site calls `navigator.permissions.revoke({name:"push", userVisibleOnly:true})`, should `navigator.permissions.query({name:"push", userVisibleOnly:false})` then return `{state:"prompt"}`? Probably so, also.

On the other hand, if `navigator.permissions.query({name:"push", userVisibleOnly:false})` returns `{state:"granted"}` and the site calls `navigator.permissions.revoke({name:"push", userVisibleOnly:false})`, should `navigator.permissions.query({name:"push", userVisibleOnly:true})` then return `{state:"prompt"}`? Probably not, since the site only asked to revoke the stronger grant.

jyasskin:
This was specified by #97's "[stronger than](https://w3c.github.io/permissions/#permissiondescriptor-stronger-than)" term.

PR #95, opened 2016-04-29
Allow UAs to maintain zero or multiple permission stores.

### LINKED ISSUE #84: Why is "retrieve the permission storage" not in "Permission store"?

### LINKED ISSUE #86: Model temporary permissions better

jyasskin:
Each origin now gets its own store, and may have several to model
tab-specific or realm-specific grants.

Fixes #84 and fixes #86.

@raymeskhoury @jan-ivar @annevk, here's an attempt to handle temporary permissions by leaving the selection of where to read and write permission-grants up to the UA. What do you think?

https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/multiple-stores/index.bs

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -49,6 +49,7 @@ spec: html
         text: parent browsing context
         text: queue a task
         text: same origin
+        text: the environment settings object's realm
         text: top-level browsing context
 spec: ui-events
     type: dfn
@@ -377,18 +378,22 @@ spec: webidl
     Permission Stores
   </h2>
   <p>
-    A <dfn export>permission store</dfn> maps {{PermissionName}}s to instances
-    of {{PermissionStorage}} or one of its subtypes. A <a>permission store</a>
-    has one or more associated <a>origin</a>s, and SHOULD have only one.
+    A <dfn export lt="permission store|permission stores">permission
+    store</dfn> maps {{PermissionName}}s to instances of
+    {{PermissionStorage}} or one of its subtypes. A <a>permission
+    store</a> has one or more associated <a>origin</a>s, and SHOULD
+    have only one.
   </p>
   <p>
-    The <a>user agent</a> MAY maintain any number of <a>permission store</a>s to
+    The <a>user agent</a> MAY maintain any number of <a>permission stores</a> to
     record what capabilities users have granted web sites permission to access.
-    The <a>user agent</a> may create new <a>permission store</a>s, for example
-    to serve a particular <a>origin</a>, <a>origin</a> within a <a>top-level
-    browsing context</a>, or <a>realm</a>, and when it does so, it may
-    initialize it using the mappings of any other <a>permission store</a>
-    associated with the <a>same origin</a>.
+    If the UA maintains any <a>permission stores</a>, the UA MUST create a new
+    <a>permission store</a> for each <a>realm</a> and may create new
+    <a>permission stores</a> for other reasons, for example to serve a
+    particular <a>origin</a> or <a>origin</a> within a <a>top-level browsing
+    context</a>. The UA may initialize new <a>permission stores</a> as empty or
+    as a copy of the mappings of any other <a>permission store</a> associated
+    with the <a>same origin</a>.
   </p>
 
   <pre class='idl'>
@@ -406,9 +411,9 @@ spec: webidl
   <ol>
     <li>Let <var>settings</var> be the <a>current settings object</a>.</li>
     <li>
-      Let <var>store</var> be some <a>permission store</a> associated with the
-      <a>same origin</a> as <var>settings</var>. This store SHOULD be chosen
-      consistently for a given <a>environment settings object</a>.
+      Let <var>store</var> be the <a>permission store</a> associated with
+      <var>settings</var>' <a lt="the environment settings object's
+      Realm">realm</a>.
     </li>
     <li>
       If <var>store</var> exists and maps <var>name</var> to a value, return
@@ -430,9 +435,9 @@ spec: webidl
     <li>
       Let <var>stores</var> be a collection of zero or more <a>permission
       store</a>s associated with the <a>same origin</a> as <var>settings</var>.
-      If <var>stores</var> is non-empty, it SHOULD include the permission store
-      used in the <a>retrieve the permission storage</a> algorithm for
-      <var>settings</var>.
+      If <var>stores</var> is non-empty, it MUST include the permission store
+      associated with <var>settings</var>' <a lt="the environment settings
+      object's Realm">realm</a>.
     </li>
     <li>
       Write a mapping from <var>name</var> to <var>storage</var> into each
--- a/index.bs
+++ b/index.bs
@@ -25,6 +25,8 @@ Markup Shorthands: css no, markdown yes
 </pre>
 <pre class="anchors">
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
+    type: dfn
+        text: realm; url: sec-code-realms
     type: interface
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
@@ -39,12 +41,14 @@ spec: promises-guide; urlPrefix: https://www.w3.org/2001/tag/doc/promises-guide#
 spec: html
     type: dfn
         text: browsing context
+        text: current settings object
         text: environment settings object
         text: event handler
         text: event handler event type
         text: origin
         text: parent browsing context
         text: queue a task
+        text: same origin
         text: top-level browsing context
 spec: ui-events
     type: dfn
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -405,38 +400,66 @@ spec: webidl
...
   </p>
   <ol>
-    <li>If the <a>user agent</a> has a {{PermissionStorage}} associated
-    with the <a>permission storage identifier</a> in its permission store,
-    it MUST return the {{PermissionStorage}}.
+    <li>Let <var>settings</var> be the <a>current settings object</a>.</li>
+    <li>
+      Let <var>store</var> be some <a>permission store</a> associated with the
+      <a>same origin</a> as <var>settings</var>. This store SHOULD be chosen
+      consistently for a given <a>environment settings object</a>.
```
martinthomson: Would it be possible to simply state that the browser picks the permission store associated with the current realm on the understanding that this is the finest granularity we support?

martinthomson: I say that on the presumption that there is only one permission store that is relevant to the request and that there is always one: either an immutable default, or one at any of the many different scopes that stores might appear.

index.bs
```diff
@@ -405,38 +400,66 @@ spec: webidl
...
-    <li>Otherwise, it MUST write the new {{PermissionStorage}} to its
-    permission store.
+    <li>
+      Write a mapping from <var>name</var> to <var>storage</var> into each
+      <a>permission store</a> in <var>stores</var>, overwriting any existing
+      mapping with the same <var>name</var>.
+
+      Issue(91): This isn't quite sufficient for complex {{PermissionStorage}}
+      subclasses, which may need to modify the old storage instead of just
+      overwriting it.
```
martinthomson: The rules for which subset of the permissions stores that the browser will actually write to are browser-specific, as we've established.  This algorithm doesn't really enable that.  It says "find all that match", then "write too all of them".  It needs to include a second step that is "identify which of the stores that the user agent wants to persist the change to".

martinthomson:
The storage model is a little wonky here.  The key to the store is the `name`, rather than the name and the full descriptor.  That means I can't give a site access to one specific camera, I have to store permissions for ALL cameras.  I'm sure that's just a mistake, or it's being addressed elsewhere, but it seems like a very important condition.

index.bs
```diff
@@ -369,33 +373,24 @@ spec: webidl
...
+    Permission Stores
   </h2>
   <p>
-    <a lt='user agent'>User agents</a> MAY use a form of storage to
-    keep track of web site permissions. When they do, they MUST have a
-    <dfn export>permission storage identifier</dfn> which is linked to a
-    {{PermissionStorage}} instance or one of its subtypes.
+    A <dfn export>permission store</dfn> maps {{PermissionName}}s to instances
+    of {{PermissionStorage}} or one of its subtypes. A <a>permission store</a>
+    has one or more associated <a>origin</a>s, and SHOULD have only one.
```
annevk: The last sentence could use some elaboration.

index.bs
```diff
@@ -369,33 +373,24 @@ spec: webidl
...
-    object</a> <var>settings</var>, the UA MUST return a tuple consisting
-    of:
+    The <a>user agent</a> MAY maintain any number of <a>permission store</a>s to
+    record what capabilities users have granted web sites permission to access.
+    The <a>user agent</a> may create new <a>permission store</a>s, for example
+    to serve a particular <a>origin</a>, <a>origin</a> within a <a>top-level
+    browsing context</a>, or <a>realm</a>, and when it does so, it may
+    initialize it using the mappings of any other <a>permission store</a>
+    associated with the <a>same origin</a>.
   </p>
```
annevk: Is this really at the granularity of a store or an individual permission? E.g., "persistent-storage" would never be for a single top-level browsing context.

jyasskin: Oh, fun. You'd require that "persistent-storage" only has global/persistent permissions, even in Firefox?

index.bs
jyasskin: I've added an example of the domain-suffix grants. Is that what you meant?

index.bs
jyasskin: Oh, I meant "Let stores be a collection of zero or more permission stores associated with the same origin as settings." to allow the UA to pick whichever subset it wants, possibly differently for each operation.

index.bs
jyasskin: Yeah, I think that works. I shied away from it because @annevk didn't want to have per-realm stores at all, but once we allow them, requiring them in the spec doesn't actually require implementations to store them separately.

jyasskin:
> The storage model is a little wonky here. The key to the store is the name, rather than the name and the full descriptor. That means I can't give a site access to one specific camera, I have to store permissions for ALL cameras. I'm sure that's just a mistake, or it's being addressed elsewhere, but it seems like a very important condition.

@martinthomson It's being addressed in #91: we can't use the descriptor as the key because descriptors are information on searching for a particular capability, rather than the capability itself. For example, [`MediaTrackConstraints`](https://w3c.github.io/mediacapture-main/#idl-def-MediaTrackConstraints) has all sorts of attributes of the camera or mic, but we'd only store the camera or mic id in the permission store. "push" and "midi" (#90), on the other hand, have a lattice structure on their permissions, where requesting `{name:'push',userVisibleOnly:false}` should imply getting `{name:'push',userVisibleOnly:true}`, and again storing the descriptor would do the wrong thing.

index.bs
jan-ivar: ?? Firefox only supports persistent permissions per-origin, right? Or does "persistent-storage" mean something else?

index.bs
martinthomson: This wouldn't necessitate a per-realm store, just that every real has a single store that is authoritative for that realm.

Or is that really the case?  Or is it that there is a hierarchy of stores for a given realm and that looking for a permission requires checking each in turn?  If the realm has a store created by storing permissions for permission "A" and it queries for permission "B", which is not yet stored in the realm.  It could have a permission for "B" stored in a per-origin store, though.  This all suggests that you either have a chain of stores to check, or that you have a store for a realm that might be initially populated with information from other stores.

jan-ivar: @martinthomson With an initially populated realm-store, wouldn't it need to update if the end-user grants persistent permission to its origin, and does so from, say, another tab or through some central UX?

martinthomson: Right.  That implies to me that a synthesis is needed.

Wow, this got complex pretty fast.

jyasskin: The current PR gives UAs enough flexibility to do this right:
> The UA may initialize new <a>permission stores</a> as empty or as a copy of the mappings of any other <a>permission store</a> associated with the <a>same origin</a>.

and 

> Let <var>stores</var> be a UA-selected subset of the <a>permission stores</a> … Write a mapping … into each <a>permission store</a> in <var>stores</var>

The UA would initialize realm-store/"B" from origin-store/"B", and if "B" is updated globally from another realm, it would just select the subset as all realm-stores + the origin-store.

I can drop the requirement for a per-realm store and just replace the "SHOULD be chosen consistently" from my original text with MUST, which will probably simplify it a bit.

index.bs
annevk: I did mean per-origin. I just meant that several top-level browsing contexts of the same origin would share the same permission. Though maybe not if some other setting was in play, like privacy mode, but we generally haven't defined or acknowledged privacy mode in standards.

jan-ivar: MediaCapture and Streams [now mandates](https://github.com/w3c/mediacapture-main/commit/bbdee2433051107e597a4fbd277b6d619113ac54) that any top-level browsing context that already has at least one un-stopped MediaStreamTrack in it must be allowed to request new tracks from the same device without a prompt. If this is the sort of thing we're trying to codify, then I imagine this tab might see (for the purposes of this discussion) `'allowed temporarily'`, while other tabs from the same origin see `'prompt'`. The view here would be that the first tab has implicitly been granted some temporary privileges that the other same-origin tabs have not (not sure how we'd describe that this is limited to a certain device, though most users have just one camera/mic, so maybe that's not super important?)

This suggests to me at least two stores: a per-origin-store for persistent permissions, and a top-level-browsing-context-store for temporary ones, and some rules for when changes to the former affects the latter. Are there any examples that need more than two stores?

jyasskin:

PR #96, opened 2016-05-11
Describe the permission store using constraints instead of a full model.

### LINKED ISSUE #84: Why is "retrieve the permission storage" not in "Permission store"?

### LINKED ISSUE #86: Model temporary permissions better

jyasskin:
Here's a third attempt at modeling the permission store, as an alternative to #95, based on @raymeskhoury's suggestions.

This says that each realm has a permission store that's a set of entries for each capability, or a flag saying that capability is denied. The UA gets to ignore writes if it wants and can post tasks to update a realm's store when it ["receives new information about the user’s intent"](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/set-permission-store/index.bs#relevant-permission-change), but each particular capability can add constraints on these updates.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/set-permission-store/index.bs#permission-stores

@alvestrand, this removes the "request a permission" entry point, because it's now straightforward to read and write the current permission store directly. We might want to add a "prompt the user to choose between several options" to match what you're using in [`getUserMedia()`](http://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia) step 7. The old algorithm didn't return a `MediaStream` anyway as step 8 expects. I think that should be a separate change.

@martinthomson @annevk

Fixes #84 and fixes #86.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -27,6 +27,7 @@ Markup Shorthands: css no, markdown yes
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
     type: dfn
         text: Realm; url: sec-code-realms
+        text: current realm; url: current-realm
     type: interface
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
@@ -37,12 +38,14 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
 spec: html
     type: dfn
         text: browsing context
+        text: current settings object
         text: environment settings object
         text: event handler
         text: event handler event type
         text: origin
         text: parent browsing context
         text: queue a task
+        text: the realm's settings object
         text: top-level browsing context
 spec: ui-events
     type: dfn
@@ -371,17 +374,25 @@ spec: webidl
   <p>
     Each <a>realm</a> has a <dfn export>permission store</dfn> consisting of,
     for each {{PermissionName}}, either a <dfn><i>denied flag</i></dfn> or a set
-    of instances of that {{PermissionName}}'s <a>permission grant type</a>. The
-    UA may initialize and update a <a>realm</a>'s <a>permission store</a> almost
-    arbitrarily, subject to the constraints in <a
-    href="#permission-store-constraints"></a>.
+    of <dfn>permission entries</dfn>, instances of that {{PermissionName}}'s
+    <a>permission entry type</a>, which will be {{PermissionStorageEntry}} or
+    one of its subtypes. The <a>permission entry type</a> specifies the
+    <a>equivalent permission entry</a> relation used to de-duplicate the
+    <a>permission entries</a>.
+  </p>
 
-    Each web API that relies on user permission defines a <a>permission storage
-    type</a> which is {{PermissionStorageEntry}} or one of its subtypes. The UA
-    provides <a>read</a>, <a>write</a>, and <a>revoke</a> abstract operations
-    that let web APIs interact with the user's choices about how permission
-    decisions are remembered. The UA is <em>not</em> required to persist
-    permission decisions in any way.
+  <p>
+    The UA may initialize <a>realm</a>'s <a>permission store</a> almost
+    arbitrarily, subject to the constraints in <a
+    href="#permission-store-constraints"></a>. The UA may also <a>queue a
+    task</a> to update a <a>realm's</a> <a>permission store</a> almost
+    arbitrarily, also subject to the constraints in <a
+    href="#permission-store-constraints"></a>.
+  </p>
+
+  <p>
+    The UA may store <a>permission entries</a> more persistently than a
+    <a>realm</a>, but this specification doesn't constrain how.
   </p>
 
   <pre class='idl'>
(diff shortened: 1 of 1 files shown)


alvestrand:
@jyasskin your PermissionStorageEntry stores permission names. This won't do - it has to store permission descriptions. Otherwise, it's impossible to store "you have permission to access microphone A, but not microphone B".

I didn't read any further than this.

alvestrand:
A "realm" is a concept I can't remember seeing referenced normatively in a W3C DOM-related spec before. I suspect that you are thinking of something completely specific that links in some way to top level browsing context, but it would be better to say which one you're thinking of.

jyasskin:
@alvestrand "Permission entries for a PermissionName are instances of that PermissionName's [permission entry type](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/set-permission-store/index.bs#permission-entry-type), which will be PermissionStorageEntry [which is just a name] or **one of its subtypes**." It's up to the 'microphone' spec to define what data it needs in order to identify a microphone. You'll be able to look at the [Bluetooth spec](https://webbluetoothcg.github.io/web-bluetooth/#dictdef-bluetoothpermissionstorage) for how to do this, although it's not updated for this patch yet.

Realms are used in https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects, and increasingly across the whole HTML spec. They're 1-1 with global objects and settings objects, which are per-frame rather than per-top-level-browsing-context. Because not all browsers and permissions work by top-level-browsing-context, I can't tie the base spec to those, but one of the extra constraints the media specs add can be that an iframe's realm should only get permission entries that were added by other realms with the same origin & top-level-origin.

index.bs
```diff
@@ -365,76 +380,123 @@ spec: webidl
...
+
   <p>
-    To <dfn>get a permission storage identifier</dfn> for a
-    {{PermissionName}} <var>name</var> and an <a>environment settings
-    object</a> <var>settings</var>, the UA MUST return a tuple consisting
-    of:
+    The UA may initialize <a>realm</a>'s <a>permission store</a> almost
+    arbitrarily, subject to the constraints in <a
+    href="#permission-store-constraints"></a>. The UA may also <a>queue a
+    task</a> to update a <a>realm's</a> <a>permission store</a> almost
```
jyasskin: Making the UA queue tasks to update the realm's permission store gives a guarantee that it doesn't change within a single turn/microtask checkpoint. However, that's not so valuable since functions like `query()` and `getCurrentLocation()` operate in parallel and so could be affected by a task posted while they're executing.

@raymeskhoury has suggested that we fall back to having *read* and *write* return completely UA-determined results. I think that's plausible. It would make it harder to write things like "the permission store only changes if the UA receives new information about the user's intent" here, but I think (e.g. in the media capture spec) "the UA must not return cameras entries from 'read the permission store' unless the user has explicitly given permission for the origin to access them" still works.

What do folks think?

annevk:
Instead of realm, I think it would be better to put this on either the settings object or the global object. Let's leave realms to JavaScript.  We only use realms in HTML to define where the settings objects live and where the global object comes from.

jyasskin:
OK, that's easy. Thanks for the direction @annevk. So you like the rest of the structure? Do you have any other constraints UAs should follow?

index.bs
```diff
@@ -365,76 +380,123 @@ spec: webidl
...
-    <li>optional UA-specific data like whether <var>settings</var>'
-    <a>browsing context</a> has a <a>parent browsing context</a>, or
-    <var>settings</var>' <a>top-level browsing context</a>'s <a>origin</a>
-    </li>
-  </ol>
+
+  <p>
+    The UA may store <a>permission entries</a> more persistently than a
+    <a>realm</a>, but this specification doesn't constrain how.
+  </p>
```
annevk: It should at least be constrained to origins. Cannot allow new kind of cookies. And ideally we do make it possible for certain permissions to require something more here. E.g., persistent storage isn't really useful without origin-wide permission.

index.bs
```diff
@@ -365,76 +380,123 @@ spec: webidl
...
+      origin</a> as one of the |realms|.
+    </p>
+
+    <p class="issue" id="issue-same-domain-permissions">
+      It's not clear if some current browsers share permissions more widely than
+      a single origin. For example, when a user grants permission for
+      https://foo.com/ to use a capability, some browsers may also give access
+      to any origin with a domain ending in ".foo.com". This specification may
+      need to allow this.
+    </p>
```
annevk: It seems like a very bad idea to cross schemes. Domains might be okay, but that also very much depends.

martinthomson: When cookies did it, it was a bad idea. I am very much opposed to an endorsement of anything that has a scope larger than scheme and host. I would prefer to retain scheme,  host, and port if possible.

jyasskin:

PR #97, opened 2016-05-25
Give the UA permission to return anything it wants from query().

jyasskin:
This option is the most straightforward about giving the UA nearly complete flexibility. The main constraint is that the result should only change when the UA gets new information about the user's intent, but we don't say what constitutes new information. This version also gives some examples ([media](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/ua-can-return-anything/index.bs#media-devices) and [durable storage](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/ua-can-return-anything/index.bs#persistent-storage)) of constraining permission behavior further on a per-API basis.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/ua-can-return-anything/index.bs

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -32,10 +32,13 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: dfn
+        text: ancestor browsing context
         text: origin; url: browsers.html#origin-2
 spec: mediacapture-main; urlPrefix: https://w3c.github.io/mediacapture-main/#
     type: attribute
         for: MediaDeviceInfo; text: deviceId
+    type: element-attr
+        for: iframe; text: allowusermedia; url: iframe-allowusermedia
     type: method
         for: MediaDevices; text: getUserMedia(); url: dom-mediadevices-getusermedia
 </pre>
@@ -43,6 +46,7 @@ spec: mediacapture-main; urlPrefix: https://w3c.github.io/mediacapture-main/#
 spec: html
     type: dfn
         text: browsing context
+        text: browsing context container
         text: current settings object
         text: environment settings object
         text: event handler
@@ -51,6 +55,7 @@ spec: html
         text: origin
         text: parent browsing context
         text: queue a task
+        text: responsible browsing context
         text: responsible event loop
         text: the realm's settings object
         text: top-level browsing context
@@ -313,6 +318,7 @@ spec: webidl
       "device-info",
       "background-sync",
       "bluetooth",
+      "persistent-storage",
     };
   </pre>
   <p>
@@ -477,6 +483,14 @@ spec: webidl
       permissions are associated with permission to use media devices as
       specified in [[GETUSERMEDIA]] and [[audio-output]].
     </p>
+    <p>
+      If the <a>current settings object</a>'s <a>responsible browsing
+      context</a> or any of its <a>ancestor browsing contexts</a> has a
+      <a>browsing context container</a> that isn't an <{iframe}> element with
+      the <{iframe/allowusermedia}> attribute specified, then the <a>permission
+      state</a> of any descriptor with a {{PermissionDescriptor/name}} of
+      {{"camera"}} or {{"microphone"}} must be {{"denied"}}.
+    </p>
     <dl>
       <dt>
         <a>permission descriptor type</a>
@@ -569,6 +583,25 @@ spec: webidl
       The <dfn for="PermissionName" enum-value>"background-sync"</dfn>
       permission is the permission associated with the usage of
       [[web-background-sync]].
+    </p>
+  </section>
+  <section>
+    <h3 id="persistent-storage">
+      Persistent Storage
+    </h3>
+    <p>
+      The <dfn for="PermissionName" enum-value>"persistent-storage"</dfn>
+      permission allows an origin to make its <a>site storage unit</a> contain a
+      <a>persistent box</a>. {{"persistent-storage"}} is a <a>boolean
+      feature</a>.
+    </p>
+    <p>
+      If a realm with <a>origin</a> |O| <a lt="request permission to
+      use">requests permission to use</a> `{name: "persistent-storage"}` and
+      that algorithm returns {{"granted"}}, then `{name:
+      "persistent-storage"}`'s <a>permission state</a> must be {{"granted"}} in
+      all realms with origin |O| until the UA receives new information about the
+      user's intent.
     </p>
   </section>
 </section>
--- a/index.bs
+++ b/index.bs
@@ -117,6 +117,26 @@ spec: webidl
   </p>
 </section>
 <section>
+  <h2 id="definitions">Definitions</h2>
+
+  <dl>
+    <dt><dfn>New information about the user's intent</dfn></dt>
+    <dd>
+      The UA may collect information about a user's intentions in any way its
+      authors believe is appropriate. This information can come from explicit
+      user action, aggregate behavior of both the relevant user and other users,
+      or other sources this specification hasn't anticipated.
+    </dd>
+
+    <dt><dfn export>Powerful feature</dfn></dt>
+    <dd>
+      A feature of a UA that some code might not be allowed to access, for
+      example because its <a>environment settings object</a> doesn't satisfy
+      some criteria, or because the user hasn't given permission.
+    </dd>
+  </dl>
+</section>
+<section>
   <h2 id="permission-descriptor">
     Descriptions of permission requests
   </h2>
@@ -126,10 +146,10 @@ spec: webidl
     };
   </pre>
   <p>
-    Each powerful feature has one or more aspects that websites may request
-    permission to access. To describe these aspects, each feature defines a
-    subtype of {{PermissionDescriptor}} to be its <a>permission descriptor
-    type</a>.
+    Each <a>powerful feature</a> has one or more aspects that websites can
+    request permission to access. To describe these requests, each feature
+    defines a subtype of {{PermissionDescriptor}} to be its <a>permission
+    descriptor type</a>.
   </p>
 
   <div class="example" id="example-descriptors">
@@ -138,35 +158,40 @@ spec: webidl
       access to system exclusive messages. Thus, its permission descriptor type
       is:
     </p>
-    <pre highlight="idl">
+    <pre>
       dictionary MidiPermissionDescriptor : PermissionDescriptor {
         boolean sysex = false;
       };
     </pre>
     <p>
-      The {{"camera"}} feature lets sites access whatever cameras are connected
-      to the user's device. Thus, its descriptor type is:
+      The {{"bluetooth"}} feature lets sites request to access whatever
+      Bluetooth devices are close to to the user's device. Thus, its descriptor
+      type is:
     </p>
-    <pre highlight="idl">
-      dictionary DevicePermissionDescriptor : PermissionDescriptor {
+    <pre>
+      dictionary BluetoothPermissionDescriptor : PermissionDescriptor {
         DOMString deviceId;
+        sequence&lt;<a idl>BluetoothRequestDeviceFilter</a>> filters;
+        sequence&lt;<a idl>BluetoothServiceUUID</a>> optionalServices = [];
       };
     </pre>
     <p>
-      General access to cameras is represented by `{name: 'camera'}`, while
-      access to a particular camera is represented by `{name: 'camera',
-      deviceId: cameraId}`.
+      General access to Bluetooth devices is represented by `{name:
+      'bluetooth'}`; access to a particular device is represented by `{name:
+      'bluetooth', deviceId: "id"}`; and access to a device with a particular
+      service is represented by `{name: 'bluetooth', filters: [{services:
+      ['service']}]}`
     </p>
   </div>
 </section>
 <section>
   <h2 id="permission-operations">Permission states</h2>
   <p>
-    The user agent is responsible for tracking what powerful features each
-    <a>realm</a> has the user's permission to use. Other specifications can use
-    the operations defined in this section to retrieve the UA's notion of what
-    permissions are granted or denied, and to ask the user to grant or deny more
-    permissions.
+    The user agent is responsible for tracking what <a>powerful features</a>
+    each <a>realm</a> has the user's permission to use. Other specifications can
+    use the operations defined in this section to retrieve the UA's notion of
+    what permissions are granted or denied, and to ask the user to grant or deny
+    more permissions.
   </p>
 
   <p>
@@ -176,22 +201,21 @@ spec: webidl
 
   <p>
     Within this section, |descriptor| is an instance of the <a>permission
-    descriptor type</a> of the powerful feature named by
+    descriptor type</a> of the <a>powerful feature</a> named by
     <code>|descriptor|.{{PermissionDescriptor/name}}</code>.
   </p>
 
   <section>
     <h3 id="reading-current-states">Reading the current permission state</h3>
     <p>
-      |descriptor|'s <dfn export>permission state</dfn> is one
-      of {{"granted"}}, {{"prompt"}}, or {{"denied"}}, indicating respectively
-      if the calling algorithm should succeed without prompting the user, show
-      the user a prompt to decide whether to succeed, or fail without prompting
-      the user. The UA must return whichever of these values most accurately
-      reflects the user's intent. Successive uses of |descriptor|'s <a>permission
-      state</a> with the same <a>current settings object</a> should return the
-      same value, unless the UA receives new information about the user's
-      intent.
+      |descriptor|'s <dfn export>permission state</dfn> is one of {{"granted"}},
+      {{"prompt"}}, or {{"denied"}}, indicating respectively if the calling
+      algorithm should succeed without prompting the user, show the user a
+      prompt to decide whether to succeed, or fail without prompting the user.
+      The UA must return whichever of these values most accurately reflects the
+      user's intent. Subsequent uses of |descriptor|'s <a>permission state</a>
+      with the same <a>current settings object</a> must return the same value,
+      unless the UA receives <a>new information about the user's intent</a>.
     </p>
 
     <p class="issue" id="issue-current-entry-incumbent-or-relevant">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
       required PermissionName name;
     };
   </pre>
   <p>
-    A permission is described by a name and other properties that depend on the
-    name. The simplest permissions require only a name, but some others have
-    more detailed structure that requires more information to describe it. In
-    that case, they should define a customized <a>permission descriptor
-    type</a> dictionary that inherits from {{PermissionDescriptor}}.
+    Each powerful feature has one or more aspects that websites may request
```
marcoscaceres: nit: link to "powerful features" dfn?

marcoscaceres: nit: please avoid using "may" without a conforming product. Maybe we can just say "websites can request"?

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+      General access to cameras is represented by `{name: 'camera'}`, while
+      access to a particular camera is represented by `{name: 'camera',
+      deviceId: cameraId}`.
+    </p>
+  </div>
+</section>
+<section>
+  <h2 id="permission-operations">Permission states</h2>
+  <p>
+    The user agent is responsible for tracking what powerful features each
```
marcoscaceres: Nit: link to "powerful features"\g ?

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+    <a>realm</a> has the user's permission to use. Other specifications can use
```
marcoscaceres: I might be mistaken, but I was under the impression we were going to avoid "realms" in web specs and leave it to ES. I wonder if there is something we could use from HTML regarding security contexts?

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+    the operations defined in this section to retrieve the UA's notion of what
+    permissions are granted or denied, and to ask the user to grant or deny more
+    permissions.
+  </p>
+
+  <p>
+    Other specifications can also add more constraints on the UA's behavior in
+    these algorithms.
```
marcoscaceres: This doesn't seem great to me... it reads like "another spec can modify these algorithms" (e.g., via monkey patch). I don't think that is the intent.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+    Within this section, |descriptor| is an instance of the <a>permission
+    descriptor type</a> of the powerful feature named by
+    <code>|descriptor|.{{PermissionDescriptor/name}}</code>.
+  </p>
+
+  <section>
+    <h3 id="reading-current-states">Reading the current permission state</h3>
+    <p>
+      |descriptor|'s <dfn export>permission state</dfn> is one
+      of {{"granted"}}, {{"prompt"}}, or {{"denied"}}, indicating respectively
```
marcoscaceres: maybe link to the enum itself? Otherwise, we will need to duplicate these values if we need to update the enum. If these are the formal definitions for each of these states, then we should break them up into a `dl`.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+      if the calling algorithm should succeed without prompting the user, show
+      the user a prompt to decide whether to succeed, or fail without prompting
+      the user. The UA must return whichever of these values most accurately
+      reflects the user's intent. Successive uses of |descriptor|'s <a>permission
```
marcoscaceres: It's unclear to me what "Successive uses" means here.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+      state</a> with the same <a>current settings object</a> should return the
```
marcoscaceres: don't you mean "MUST return the same value"... it would be strange if it suddenly changed for whatever reason. The next statement should clarify about changing the value.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+      Safari is the only known UA that returns different results from this
+      algorithm for different settings objects with the same origin. We should
+      test which of the <a
+      href="https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects">several
+      possible settings objects</a> it uses.
+    </p>
+
+    <p>
+      Some powerful features have more information associated with them than
+      just a {{PermissionState}}. For example, {{MediaDevices/getUserMedia()}}
```
marcoscaceres: The example here seems to mirror the information that was already given in the example above, no?

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+      needs to determine <em>which</em> cameras the user has granted the
+      <a>current realm</a> permission to access. Each of these features defines
+      an <a>extra permission data type</a>, and then a {{PermissionName}} |name|'s
+      <dfn export>extra permission data</dfn> is the instance of that type
+      matching the UA's impression of the user's intent. Successive uses of
+      |name|'s <a>extra permission data</a> should return the same value, unless
+      the UA receives new information about the user's intent.
```
marcoscaceres: the text " unless the UA receives new information about the user's intent" seems to be repeated throughout. Can we capture this as a concept to avoid having to repeat it?

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+    </p>
+  </section>
+  <section>
+    <h3 id="requesting-more-permission">Requesting more permission</h3>
+
+    <div algorithm="request-permission-to-use">
+      <p>
+        To <dfn export>request permission to use</dfn> a |descriptor|, the UA
+        must perform the following steps. Note that these steps may wait for
```
marcoscaceres: the "note" text should be moved into it's own "note" section.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+        the main thread. This algorithm returns either {{"granted"}} or
+        {{"denied"}}.
+      </p>
+      <ol>
+        <li>
+          If |descriptor|'s <a>permission state</a> is not {{"prompt"}}, return
+          that value and abort these steps.
+        </li>
+        <li>
+          Show the user a prompt asking their permission for the calling
```
marcoscaceres: Note: "show" implies visual prompt - some UAs may prompt through different modalities, so maybe use prompt as a verb:

> "Prompt the user for permission x..."

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+          algorithm to use powerful feature described by |descriptor|.
+        </li>
+        <li>
+          If the user grants permission, return {{"granted"}}; otherwise return
+          {{"denied"}}. Depending on the details of the user's interaction, the
+          UA may also treat this as new information about the user's intent for
+          other realms with the same origin.
```
marcoscaceres: nit: link to realms and origin.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+
+          <p class="note">
+            This is intentionally vague about the details of the permission UI
+            and how the UA infers user intent. UAs should be able to explore
+            lots of UI within this framework.
```
marcoscaceres: 👍

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+          </p>
+        </li>
+      </ol>
+    </div>
+
+    <div algorithm="prompt-user-to-choose">
+      <p>
+        To <dfn export>prompt the user to choose</dfn> one of several options
+        associated with a |descriptor|, the UA must perform the following steps.
+        Note that these steps may wait for user input, so they should not be
```
marcoscaceres: Nit: if possible, we should avoid duplicating this text. Maybe we can move this up as a general note about prompting.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+            and how the UA infers user intent. UAs should be able to explore
+            lots of UI within this framework.
+          </p>
+        </li>
+      </ol>
+    </div>
+
+    <div algorithm="prompt-user-to-choose">
+      <p>
+        To <dfn export>prompt the user to choose</dfn> one of several options
```
marcoscaceres: Link to options? Particularly as they go in as input to the algorithm.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+        <li>
+          If |descriptor|'s <a>permission state</a> is {{"denied"}},
+          return {{"denied"}} and abort these steps.
+        </li>
+        <li>
+          If |descriptor|'s <a>permission state</a> is {{"granted"}}, the UA may
+          return one of the options and abort these steps. If the UA returns
+          without prompting, then successive <a lt="prompt the user to
+          choose">prompts for the user to choose</a> from the same set of
+          options with the same |descriptor| should return the same option,
```
marcoscaceres: As above. I think this "should" needs to be a MUST, whereby the permission state could be modified out-of-band (e.g., the user changes the permission via browser UI)... in which case, it MUST change to reflect the new information asynchronously. However, this reads like the permission could be synchronously changed before the turn of the event loop, which I don't think is the intent (maybe it doesn't matter, but I don't think we would want to implement it that way because there will be IPC involved).

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+        <li>
+          Show the user a prompt asking them to choose one of the options or
+          deny permission, and wait for them to choose. If the calling algorithm
+          specified extra information to show in the prompt, show it.
+        </li>
+        <li>
+          If the user chose an option, return it; otherwise return {{"denied"}}.
+          Depending on the details of the user's interaction, the UA may also
+          treat this as new information about the user's intent for other realms
+          with the same origin.
```
marcoscaceres: So, given what is written here, it would cause a propagation across realms. It sounds like "queue a task" might be needed for propagating the change, no?

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+
+          <p class="note">
```
marcoscaceres: Again, we should avoid duplicating text. Let's centralize this kind of text somewhere.

index.bs
```diff
@@ -101,21 +118,190 @@ spec: webidl
...
+          </p>
+        </li>
+      </ol>
+    </div>
+  </section>
+  <section>
+    <h3 id="reacting-to-revocation">Reacting to users revoking permission</h3>
+
+    <p>
+      When the UA learns that the user no longer intends to grant permission for
```
marcoscaceres: ❤️"Well hello queue a task" ❤️ ... The algorithms above need to link to here.

index.bs
```diff
@@ -132,27 +318,39 @@ spec: webidl
...
     </dt>
     <dd>
-      {{PermissionDescriptor}} or one of its subtypes.
-      If unspecified, this defaults to {{PermissionDescriptor}}.
+      <p>
+        {{PermissionDescriptor}} or one of its subtypes.
+        If unspecified, this defaults to {{PermissionDescriptor}}.
+      </p>
+      <p>
+        The feature can define a partial order on instances. If |descriptorA| is
```
marcoscaceres: Not sure what " partial order " means here? Maybe example is needed?

index.bs
```diff
@@ -132,27 +318,39 @@ spec: webidl
...
+        <dfn for="PermissionDescriptor">stronger</dfn> than |descriptorB|, then
+        if |descriptorA|'s <a>permission state</a> is {{"granted"}},
+        |descriptorB|'s <a>permission state</a> must also be {{"granted"}}, and
```
marcoscaceres: Nit: These are statements of fact, so please use "is" instead of "must". for instance, `<a>permission state</a> is {{"granted"}}`.

index.bs
```diff
@@ -165,45 +363,42 @@ spec: webidl
...
-      if the request can fail exceptionally. (Merely being denied
-      permission is not exceptional.) Used by {{Permissions}}'
-      {{Permissions/request()}} method, which handles
-      reading and writing the <a>permission store</a>. If unspecified, this
-      defaults to the <a>boolean permission request algorithm</a>.
+      Takes an instance of the <a>permission descriptor type</a> and a
+      newly-created instance of the <a>permission result type</a>. Uses the
+      algorithms in <a href="#requesting-more-permission"></a> to show the user
+      any necessary prompt to try to increase permissions, and updates the
+      instance <a>permission result type</a> to match. May return a {{Promise}}
```
marcoscaceres: Wait! it would be kinda sad if some callers returned a Promise an others don't. This appears to be what is implied here, or am I misreading this?

index.bs
```diff
@@ -165,45 +363,42 @@ spec: webidl
...
+      states</a> or <a>extra permission data</a>. Run by {{Permissions}}'
+      {{Permissions/revoke()}} method and run when the UA receives new
+      information about the user's intent. If unspecified, this defaults to
+      doing nothing.
     </dd>
   </dl>
   <p>
-    A <dfn export>boolean permission</dfn> is a <a>permission</a> with all types
+    A <dfn export>boolean feature</dfn> is a powerful feature with all types
     and algorithms defaulted.
```
marcoscaceres: I know this is not being changed by this PR, but this is quite vague. Do "powerful features" define defaults?

index.bs
```diff
@@ -295,23 +506,34 @@ spec: webidl
...
-          cameras, a query without the {{deviceId}} will return {{"prompt"}}.
+          cameras, a query without the {{DevicePermissionDescriptor/deviceId}}
+          will return {{"prompt"}}.
+        </p>
+        <p class="issue" id="issue-media-granted">
+          It may not make sense for `{name: "camera"}`'s <a>permission state</a>
+          to ever be {{"granted"}}: if the site calls
+          {{MediaDevices/getUserMedia()}} with constraints that none of the
+          user's devices satisfy, the UA cannot return a device without
+          prompting, but saying {{"granted"}} promised to do so.
```
marcoscaceres: I'm having a hard time parsing this paragraph. I don't understand "but saying {{"granted"}} promised to do so." Maybe change it to active voice? promised by who?

index.bs
```diff
@@ -363,78 +585,25 @@ spec: webidl
...
+    </h3>
+    <p>
+      The <dfn for="PermissionName" enum-value>"persistent-storage"</dfn>
+      permission allows an origin to make its <a>site storage unit</a> contain a
+      <a>persistent box</a>. {{"persistent-storage"}} is a <a>boolean
+      feature</a>.
+    </p>
+    <p>
+      If a realm with <a>origin</a> |O| <a lt="request permission to
+      use">requests permission to use</a> `{name: "persistent-storage"}` and
```
marcoscaceres: nit: you don't need the lt here, as it's the same as the text content.

index.bs
```diff
@@ -494,81 +641,17 @@ spec: webidl
     };
   </pre>
   <p>
-    {{PermissionStatus}} instances are created with the following
-    internal slots:
+    {{PermissionStatus}} instances are created with a 
+    <dfn for="PermissionStatus" attribute>\[[query]]</dfn> internal slot, which is an instance of a feature's <a>permission descriptor type</a>.
```
marcoscaceres: nit, line length.

marcoscaceres:
Few comments, nits, etc. Thanks for putting this PR together @jyasskin!

index.bs
jyasskin: I don't see anything in HTML about security contexts. We could make up a term to mean "the thing that code runs inside which has a global object and a settings object", but why bother if Javascript already has it?

index.bs
jyasskin: What kind of hook would read better to you? It really is the case that other specs need to be able to modify the values these algorithms can return, and you can see some examples under push, media, and persistent-storage.

index.bs
jyasskin: I don't think we can ever update the enum once it's escaped into the web. Too many people will test for two of the states and `else` the third.

Similarly, updating this list will be the least of our worries if we change an enum value: we'll also have to update all of the callers in other specs.

index.bs
jyasskin: "Subsequent"?

index.bs
jyasskin: Sure.

index.bs
jyasskin: I can define the term, but I don't immediately see a shorter phrase that works well. Do you have ideas?

index.bs
jyasskin: Changed to 'must'. I don't think constraining changes to turn boundaries is particularly useful to authors, since this algorithm's running in parallel anyway and so you can't arrange to have two calls to it happen within the same turn. I tried to guarantee this in #96 if you want to take a look.

index.bs
jyasskin: What would the task's steps be? Since information about the user's intent always comes in asynchronously, and it only affects the return values from other mostly-unconstrained algorithms, I don't think we generally need to talk about the tasks involved.

index.bs
jyasskin: This is different from the other "may treat this as new information" steps and wouldn't be appropriate as their behavior.

index.bs
jyasskin: It's the math term: https://en.wikipedia.org/wiki/Partially_ordered_set.

index.bs
jyasskin: These are actually constraints on what the UA returns from the _permission state_ algorithm.

index.bs
jyasskin: This algorithm is entirely internal, so what it returns isn't exposed to javascript. That said, I can probably have it possibly-throw exceptions instead of possibly-returning Promises.

index.bs
jyasskin: The types and algorithms above define defaults.

index.bs
jyasskin: It's "request" vs "requests", and Bikeshed's automatic pluralization only applies to the end of a term. (@tabatkins)

index.bs
jyasskin: I don't think there is a definition of "powerful features". It used to be in Secure Contexts, but it got taken out. I was going back and forth on adding a "Definitions" section in this patch, but I guess I'll do it for this one.

And I've switched to "can".

index.bs
jyasskin: Done.

index.bs
jyasskin: I actually expect that the "camera" permission descriptor will become more like [`MediaStreamConstraints`](https://w3c.github.io/mediacapture-main/#idl-def-MediaStreamConstraints), rather than just the device ID that's there now. The descriptor is about describing a desired permission, while the `extra permission data` is about describing a granted permission, and they can be totally different kinds of data.

I've updated the example above to make that more clear, by using `"bluetooth"` instead of `"camera"`.

index.bs
jyasskin: Done.

index.bs
jyasskin: Done.

index.bs
jyasskin: Done.

index.bs
jyasskin: Done.

index.bs
jyasskin: You mean `<var>` them? Done.

index.bs
jyasskin: How's this?

index.bs
jyasskin: Fixed.

index.bs
tabatkins: Correct, term fixup anywhere in the phrase would be cool but complicated.  (If this form shows up a lot, you can add it manually as a linking text to the dfn.)

martinthomson:
Yes, this is good.  You should merge this and we can iterate on the outcome.

PR #99, opened 2016-07-11
feat(travis): autopublish with echidna (closes #24)

### LINKED ISSUE #24: Move to using Echidna
Need to set up automated publishing after FPWD.

marcoscaceres:
Prevent TR from becoming more TRash.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,19 @@ language: generic
 
 script: bash ./deploy.sh
 
+branches:
+  only:
+    - master
+
 env:
   global:
   - ENCRYPTION_LABEL: 104fbe69e8fa
   - COMMIT_AUTHOR_EMAIL: travis-ci@w3.org
+  - URL="https://w3c.github.io/permissions/"
+  - DECISION="https://lists.w3.org/Archives/Public/public-webappsec/2015Mar/0170.html"
+  # Exports $TOKEN - the W3C Token needed by Echidna to auto-publish this spec
+  # See: https://github.com/w3c/echidna/wiki/How-to-use-Echidna-with-ReSpec-and-GitHub#the-token
+  - secure: SneznWMBm52ct+mFI6NAVrB8k/BjO/H/hjR1eB3tjPwF0vzpL5fnbD1mmlNmVJMTWMsSUfI917JotA+7lfwoWJuHwlq2WfzQ0ybfvtuqjo7LS22T5X8IFYmCyVNR5asbJha/LNhqUyGDsZlPBgbKK0hqTd8eWjjQTztnlSrorGM=
+
+after_success:
+  - '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && curl "https://labs.w3.org/echidna/api/request" --data "url=$URL" --data "decision=$DECISION" --data "token=$TOKEN"'


.travis.yml
```diff
@@ -2,7 +2,17 @@ language: generic
 
 script: bash ./deploy.sh
 
+branches:
+  only:
+    - gh-pages
```
jyasskin: I think we want master in the only list also, or general PRs might not cause Travis to run and check that the files compile.

.travis.yml
```diff
@@ -2,7 +2,17 @@ language: generic
...
+
 env:
   global:
   - ENCRYPTION_LABEL: 104fbe69e8fa
   - COMMIT_AUTHOR_EMAIL: travis-ci@w3.org
+  - URL="https://w3c.github.io/permissions/"
+  - DECISION="https://lists.w3.org/Archives/Public/public-webappsec/2015Mar/0170.html"
+  - secure: SneznWMBm52ct+mFI6NAVrB8k/BjO/H/hjR1eB3tjPwF0vzpL5fnbD1mmlNmVJMTWMsSUfI917JotA+7lfwoWJuHwlq2WfzQ0ybfvtuqjo7LS22T5X8IFYmCyVNR5asbJha/LNhqUyGDsZlPBgbKK0hqTd8eWjjQTztnlSrorGM=
```
jyasskin: Can you `# comment` what secrets are in this string? I assume it's the token from https://github.com/w3c/echidna/wiki/How-to-use-Echidna-with-ReSpec-and-GitHub#the-token?

.travis.yml
```diff
@@ -2,7 +2,17 @@ language: generic
...
+
+after_success:
+  - '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && curl "https://labs.w3.org/echidna/api/request" --data "url=$URL" --data "decision=$DECISION" --data "token=$TOKEN"'
```
jyasskin: Does Echidna erase the difference between an Editor's Draft and a Working Draft? I've been defending some of the content here by saying that we'll need to ask for consensus on it later, but if this is going to start auto-updating https://www.w3.org/TR/permissions/ and calling it a Working Draft, I think we're going to get some objections.

jyasskin: Per https://github.com/w3c/echidna/wiki/How-to-use-Echidna-with-ReSpec-and-GitHub#the-token and https://github.com/tabatkins/bikeshed/issues/362, we also need to add back the editor IDs to https://github.com/w3c/permissions/blob/master/index.bs#L10. I'll send you a change.

.travis.yml
jyasskin: Maybe we can get away with *only* having `master`, since `deploy.sh` will push to `gh-pages`, which causes https://w3c.github.io/permissions/ to get the new document, and then echidna will pick it up from there.

marcoscaceres: > I think we want master in the only list also, or general PRs might not cause Travis to run deploy.sh to check that the files compile.

Oh, I might be confused, but I thought we only wanted it on gh-pages to allow us to keep editing on master before deploying via gh-pages. I'll change it to master tho.

.travis.yml
marcoscaceres: Will do.

.travis.yml
marcoscaceres: > I've been defending some of the content here by saying that we'll need to ask for consensus on it later, but if this is going to start auto-updating https://www.w3.org/TR/permissions/ and calling it a Working Draft, I think we're going to get some objections.

Yeah, that's why I thought that only if something hits "gh-pages" should it call echidna. However, I think it's better to only have one version of the spec (even if the stuff is controversial) - controversial stuff should just be marked as such with an issue and a call for people to collaborate on reaching consensus... or else we risk never being able to get the spec on TR because there is likely always going to be controversy.

marcoscaceres:
@jyasskin, made the changes you suggested. Let me know how you want to proceed.

.travis.yml
jyasskin: [`deploy.sh`](https://github.com/w3c/permissions/blob/master/deploy.sh) automatically compiles `master` to `gh-pages`, so we don't have this intermediate step in this repository.

.travis.yml
```diff
@@ -2,7 +2,19 @@ language: generic
...
   - ENCRYPTION_LABEL: 104fbe69e8fa
   - COMMIT_AUTHOR_EMAIL: travis-ci@w3.org
+  - URL="https://w3c.github.io/permissions/"
+  - DECISION="https://lists.w3.org/Archives/Public/public-webappsec/2015Mar/0170.html"
+  # Exports $TOKEN - the W3C Token needed by Echidna to auto-publish this spec
+  # See: https://github.com/w3c/echidna/wiki/How-to-use-Echidna-with-ReSpec-and-GitHub#the-token
+  - secure: SneznWMBm52ct+mFI6NAVrB8k/BjO/H/hjR1eB3tjPwF0vzpL5fnbD1mmlNmVJMTWMsSUfI917JotA+7lfwoWJuHwlq2WfzQ0ybfvtuqjo7LS22T5X8IFYmCyVNR5asbJha/LNhqUyGDsZlPBgbKK0hqTd8eWjjQTztnlSrorGM=
+
+after_success:
+  - '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && curl "https://labs.w3.org/echidna/api/request" --data "url=$URL" --data "decision=$DECISION" --data "token=$TOKEN"'
```
jyasskin: We could use the condition from https://github.com/w3c/permissions/blob/master/deploy.sh#L8 (`[ "$TRAVIS_PULL_REQUEST" != "false" -o "$TRAVIS_BRANCH" != "master" ]`) to allow Travis to run on all branches, but only ping Echidna for master commits.

LGTM either way.

jyasskin:
Nobody objected to @annevk's suggestion to publish at https://lists.w3.org/Archives/Public/public-webappsec/2016Aug/0042.html, so I'm happy with this PR. Please merge when you're ready.

PR #100, opened 2016-07-11
Editorial fixes

marcoscaceres:


index.bs
```diff
@@ -73,47 +73,42 @@ spec: webidl
...
-    handle permissions on the Web platform. Web APIs have different ways to
-    deal with permissions. The [[notifications]] API allows developers to
+    This document specifies an API to query and request permissions to
+    powerful features on the Web platform. Web APIs have different ways to
+    deal with permissions. The [[Notifications]] API allows developers to
     request a permission and check the permission status explicitly. Others
-    might only expose the status to web pages when they try to use the API,
-    like the [[geolocation-API]] which fails if the permission was not
+    expose the status to web pages when they try to use the API,
+    like the [[Geolocation-API]] which fails if the permission was not
```
jyasskin: Bikeshed wants you to change the `[[geolocation-API]]` in #geolocation too.

index.bs
```diff
@@ -73,47 +73,42 @@ spec: webidl
   </h2>
   <p><em>This section is non-normative.</em></p>
   <p>
-    This document's goal is to specify an API that will help developers to
-    handle permissions on the Web platform. Web APIs have different ways to
-    deal with permissions. The [[notifications]] API allows developers to
+    This document specifies an API to query and request permissions to
+    powerful features on the Web platform. Web APIs have different ways to
+    deal with permissions. The [[Notifications]] API allows developers to
```
jyasskin: And the `[[notifications]]` in #notifications.

index.bs
```diff
@@ -73,47 +73,42 @@ spec: webidl
   </h2>
   <p><em>This section is non-normative.</em></p>
   <p>
-    This document's goal is to specify an API that will help developers to
-    handle permissions on the Web platform. Web APIs have different ways to
-    deal with permissions. The [[notifications]] API allows developers to
+    This document specifies an API to query and request permissions to
```
jyasskin: The document also centralizes the definitions of how other specs request permission from people.

index.bs
```diff
@@ -73,47 +73,42 @@ spec: webidl
...
+    like the [[Geolocation-API]] which fails if the permission was not
     granted without allowing the developer to check beforehand.
   </p>
   <p>
-    Being able to know whether an API call is going to prompt is useful in
-    order to provide a good user experience. Unfortunately, more often than
-    not, those prompts can't be controlled by developers.
+    The Permissions API provides the tools that potentially allow developers to
+    improve the user experience of web applications that require
+    multiple permission requests.
```
jyasskin: How about "The Permissions API provides tools for developers to control when permission prompts are shown and to give up permissions that are no longer needed."? "tools that potentially allow" sounds like too much hedging.

index.bs
```diff
@@ -73,47 +73,42 @@ spec: webidl
...
   </p>
   <p>
-    The API specified in this document is meant to provide the tools so
-    that web applications can improve their user experience when
-    permissions are involved.
+    The solution described in this document is meant to be extensible, but isn't
+    meant to be applicable to all the current and future permissions
+    available in the web platform. Working Groups that are creating specifications
+    whose permission model don't fit in the model described in this document
```
jyasskin: Either "models don't" or "model doesn't".

index.bs
```diff
@@ -73,47 +73,42 @@ spec: webidl
...
-    but often through actually using the API which could lead to a
-    permission request UI if the permission was not already granted. Thus,
-    even though this API doesn't expose new fingerprinting data to
-    websites, it makes it easier for them to have discreet access to it.
-    Therefore, implementations are encouraged to have an option for users
-    to block (globally or selectively) the querying of permission states.
+    Implementations are encouraged to have an option for users to block (globally
+    or selectively) the querying of <a>permission states</a>: An adversary
+    could use a <a>permission state</a> as an element in creating a "fingerprint"
+    corresponding to an end-user. Although, a adversary can already determine
```
jyasskin: "Although an adversary"

index.bs
```diff
@@ -73,47 +73,42 @@ spec: webidl
...
   <p>
-    Permission states can be used as an element of fingerprinting by
-    websites. Usually websites could already have access to the information
-    but often through actually using the API which could lead to a
-    permission request UI if the permission was not already granted. Thus,
-    even though this API doesn't expose new fingerprinting data to
-    websites, it makes it easier for them to have discreet access to it.
-    Therefore, implementations are encouraged to have an option for users
-    to block (globally or selectively) the querying of permission states.
+    Implementations are encouraged to have an option for users to block (globally
```
jyasskin: I liked putting this encouragement at the end better. Start with the privacy consideration, and end with a possible mitigation.

index.bs
```diff
@@ -128,7 +123,7 @@ spec: webidl
       or other sources this specification hasn't anticipated.
     </dd>
 
-    <dt><dfn export>Powerful feature</dfn></dt>
+    <dt><dfn export lt="feature|features">Powerful feature</dfn></dt>
```
jyasskin: In order for `<a>powerful feature</a>` to work, you have to include it in the |'ed list. Also, Bikeshed will automatically strip "s" from the end of a word when linking it, so you only need `lt="feature|powerful feature"`.

index.bs
```diff
@@ -146,16 +141,15 @@ spec: webidl
     };
   </pre>
   <p>
-    Each <a>powerful feature</a> has one or more aspects that websites can
-    request permission to access. To describe these requests, each feature
-    defines a subtype of {{PermissionDescriptor}} to be its <a>permission
-    descriptor type</a>.
+    Each <a>powerful feature</a> has one or more <dfn export>aspects</dfn> that
```
jyasskin: For `export` to make sense, we'd need to be defining "aspect" for the whole web platform. Is that really what we want to do here?

index.bs
```diff
@@ -208,14 +202,23 @@ spec: webidl
...
     <p>
-      |descriptor|'s <dfn export>permission state</dfn> is one of {{"granted"}},
-      {{"prompt"}}, or {{"denied"}}, indicating respectively if the calling
-      algorithm should succeed without prompting the user, show the user a
-      prompt to decide whether to succeed, or fail without prompting the user.
-      The UA must return whichever of these values most accurately reflects the
-      user's intent. Subsequent uses of |descriptor|'s <a>permission state</a>
-      with the same <a>current settings object</a> must return the same value,
-      unless the UA receives <a>new information about the user's intent</a>.
+      |descriptor|'s <dfn export lt="permission state|state">permission state
```
jyasskin: I think you want `local-lt="state"`. Another spec using <a>state</a> shouldn't land here.

index.bs
```diff
@@ -208,14 +202,23 @@ spec: webidl
...
+    </p>
+
+    <ul>
+      <li>succeed without prompting the user ({{"granted"}}),</li>
+      <li>show the user a prompt to decide whether to succeed ({{"prompt"}}),</li>
+      <li>or fail without prompting the user ({{"denied"}}).</li>
+    </ul>
+
+    <p>
+      The UA returns whichever of these values most accurately reflects the user's
```
jyasskin: I think this should stay "must return". It's intended to constrain the UA, not just state a fact about what UAs do.

index.bs
```diff
@@ -243,8 +246,9 @@ spec: webidl
     <h3 id="requesting-more-permission">Requesting more permission</h3>
 
     <p class="note">
-      The algorithms in this section may wait for user input, so they should not
-      be used from other algorithms running on the main thread.
+      Implementers, please note that algorithms in this section can wait for
```
jyasskin: This note is actually for other spec authors.

index.bs
```diff
@@ -297,13 +305,13 @@ spec: webidl
           the UA receives <a>new information about the user's intent</a>.
         </li>
         <li>
-          Ask the user to choose one of the options or deny permission, and wait
-          for them to choose. If the calling algorithm specified extra
-          information to include in the prompt, include it.
+          Wait for the user to choose either one of the options or deny permission.
```
jyasskin: We could wait a very long time if we wait for them to choose something we haven't asked them about. ;)

index.bs
```diff
@@ -297,13 +305,13 @@ spec: webidl
...
+          If the calling algorithm specified extra information to include in the
+          prompt, include it.
         </li>
         <li>
           If the user chose an option, return it; otherwise return {{"denied"}}.
-          Depending on the details of the user's interaction, the UA may also
+          Depending on the details of the user's interaction,
```
jyasskin: Is the UA required to treat a choice as new information, or is it optional? I put "the UA may" to make it optional.

index.bs
```diff
@@ -375,8 +383,8 @@ spec: webidl
         <code>{name: {{"midi"}}, sysex: true}</code> ("midi-with-sysex") is
         <a>stronger than</a> <code>{name: {{"midi"}}, sysex: false}</code>
         ("midi-without-sysex"), so if the user denies access to
-        midi-without-sysex, the UA must also deny access to midi-with-sysex, and
-        similarly if the user grants access to midi-with-sysex, the UA must also
+        midi-without-sysex, the UA would also deny access to midi-with-sysex, and
```
jyasskin: I think there's actually an RFC 2119 "must" here, although being in an example that's derived from other rules in the spec rather than being required here.

jyasskin:
There's a bunch of bikeshed errors here. Could you run bikeshed over the file and fix them?

marcoscaceres:
Will do! Sorry, still new to bs.

index.bs
marcoscaceres: ah, boo. I thought it would support case insensitive look-up for bibref.

index.bs
marcoscaceres: Probably not. Reverting that.

index.bs
marcoscaceres: reverting.

index.bs
marcoscaceres: Maybe, then: 

>  This document specifies a model and an API to query and request permissions to powerful features on the Web platform.

index.bs
marcoscaceres: Better... but maybe, s/give up/relinquish

index.bs
marcoscaceres: heh, my "gutter grammar" 😜

index.bs
marcoscaceres: fixed.

index.bs
marcoscaceres: Ok, switched it back.

index.bs
marcoscaceres: fixed

index.bs
marcoscaceres: oh, that's a nice feature. Didn't know about that one.

index.bs
marcoscaceres: Ok, but shouldn't that be in an algorithm? If it's in one of the algorithms, then this is a statement of fact. If it's not in an algorithm, that might be a bug.

index.bs
marcoscaceres: Ok, added "Spec authors," to make that clear.

index.bs
marcoscaceres: Yeah, on reflection you are correct. Reverting.

index.bs
marcoscaceres: It doesn't feel great to have optional behavior in an algorithm if it's not conditional. Could it be phrased as an "if statement" somehow?

index.bs
marcoscaceres: Reverted them back to "must". I'm not a huge fan, but it's def more clear.

marcoscaceres:
trying to address BS fatal errors... getting strange ones like:
FATAL ERROR: Unknown metadata key "Repository". Prefix custom keys with "!".

index.bs
jyasskin: Both of them match; it's Bikeshed saying that we should use a consistent spelling of a `[[reference]]` within a given spec.

index.bs
jyasskin: LGTM.

index.bs
jyasskin: SGTM

index.bs
```diff
@@ -128,7 +123,7 @@ spec: webidl
       or other sources this specification hasn't anticipated.
     </dd>
 
-    <dt><dfn export>Powerful feature</dfn></dt>
+    <dt><dfn export lt="powerful feature|feature">Powerful feature</dfn></dt>
```
jyasskin: Missed the possibility of `local-lt` here too: Probably we don't want this spec to define "feature" for the whole world?

index.bs
```diff
@@ -208,14 +202,23 @@ spec: webidl
...
     <p>
-      |descriptor|'s <dfn export>permission state</dfn> is one of {{"granted"}},
-      {{"prompt"}}, or {{"denied"}}, indicating respectively if the calling
-      algorithm should succeed without prompting the user, show the user a
-      prompt to decide whether to succeed, or fail without prompting the user.
-      The UA must return whichever of these values most accurately reflects the
-      user's intent. Subsequent uses of |descriptor|'s <a>permission state</a>
-      with the same <a>current settings object</a> must return the same value,
-      unless the UA receives <a>new information about the user's intent</a>.
+      |descriptor|'s <dfn export lt="permission state" local-lt="state">
```
jyasskin: You should be able to drop the `lt` since `local-lt` doesn't override the content of the tag.

index.bs
jyasskin: I intended these paragraphs to be the algorithm that determines "|descriptor|'s permission state". Obviously that wasn't clear enough. Should we say "|descriptor|'s permission state is the result of the following algorithm"?

index.bs
jyasskin: We could use something like "If the user's interaction indicates they intend this choice to apply to other realms, then ..."?

We do need to let UAs vary on how they interpret choices here.

jyasskin:
> trying to address BS fatal errors... getting strange ones like:
> FATAL ERROR: Unknown metadata key "Repository". Prefix custom keys with "!".

https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/w3c/permissions/editorial_fixes/index.bs&output=both doesn't have any fatal errors, so that sounds like something in your bikeshed installation is broken. Repository is a known key, too. @tabatkins, any idea what it would be?

tabatkins:
Yeah, that'd be an old Bikeshed version. Go to your bikeshed folder and run `git pull --rebase`

index.bs
tabatkins: Yes, `<dfn export local-lt=feature>Powerful feature</dfn>` is the ideal markup here.

marcoscaceres: Ok, fixed both.

index.bs
marcoscaceres: Ok, added the text you suggested + plus also did the same where this same text occurred elsewhere (in the request-permission-to-use algo).

index.bs
marcoscaceres: Yes, I think that would help. Let me try to phrase it.

marcoscaceres:
@jyasskin, I explicitly marked "read-current-permission-state" as an algorithm. Is that ok?

marcoscaceres:
@jyasskin, see https://github.com/w3c/permissions/pull/100/commits/18f6961022532da20a3d83a6dc04aaa643ce0112

index.bs
```diff
@@ -201,46 +201,52 @@ spec: webidl
...
-      {{MediaDevices/getUserMedia()}} needs to determine <em>which</em> cameras
-      the user has granted the <a>current realm</a> permission to access. Each
-      of these features defines an <a>extra permission data type</a>, and then a
-      {{PermissionName}} |name|'s <dfn export>extra permission data</dfn> is the
-      instance of that type matching the UA's impression of the user's intent.
-      Subsequent uses of |name|'s <a>extra permission data</a> must return the
-      same value, unless the UA receives <a>new information about the user's
-      intent</a>.
-    </p>
+      <p>
```
jyasskin: This is a different algorithm, defining the result of

> `"permission-name"`'s extra permission data

index.bs
```diff
@@ -201,46 +201,52 @@ spec: webidl
...
   <section>
     <h3 id="reading-current-states">Reading the current permission state</h3>
-    <p>
-      |descriptor|'s <dfn export local-lt="state">permission state</dfn> is one
-      of {{"granted"}}, {{"prompt"}}, or {{"denied"}}, indicating respectively
-      if the calling algorithm should:
-    </p>
+    <div algorithm="read-current-permission-state">
+      <p>
+        To <dfn export>read the current permission state</dfn>, use the
```
jyasskin: Can we use something like `|descriptor|'s <dfn export local-lt="state">permission state</dfn> is the result of the following algorithm.`? We *could* define two phrases that mean the same thing, but it seems redundant. I defined "permission state" as the name because @annevk wanted a really concise way to get to this value, in #74 request 2.

index.bs
```diff
@@ -255,18 +265,23 @@ spec: webidl
       </p>
       <ol>
         <li>
-          If |descriptor|'s <a>permission state</a> is not {{"prompt"}}, return
-          that value and abort these steps.
+          Let <var>current state</var> be the |descriptor|'s <a>permission
```
jyasskin: FYI, you can replace `<var>foo</var>` with `|foo|`.

jyasskin:
Ok, everything except the details of 18f6961022532da20a3d83a6dc04aaa643ce0112 looks good.

index.bs
marcoscaceres: Sorry, I don't understand this comment :( 

In case I don't hear back from you in the next few hours; I'm about to shift focus to another project for 2 weeks, so could I kindly request that you fix this for me in the `editorial_fixes` branch? I've already fixed the `<var>`s (there are ~30 vars  that could maybe be converted also, but I guess we should do that in a new PR).

marcoscaceres:
rebased.

index.bs
marcoscaceres: Ok, I used your wording here (see https://github.com/w3c/permissions/pull/100/commits/64cc49aac53bc240b0a2af252a1085fa448fc1db)

index.bs
jyasskin: Oh yeah, totally. I'd forgotten that this branch was in the main repo rather than a personal fork. I'll take it over.

marcoscaceres:
(don't forget to merge me or there will be much bit rot 👍 )

PR #101, opened 2016-07-11
chore(examples): Modernize JS examples

marcoscaceres:
Make the examples a little more 2015... for the millennials.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -912,16 +912,19 @@ spec: webidl
     add the feature.
   </p>
   <pre class='highlight'>
-    &lt;script&gt;
-      navigator.permissions.query({name:'geolocation'}).then(function(result) {
-        if (result.state == 'granted') {
+    navigator.permissions.query({ name: "geolocation" }).then(({ state }) => {
+      switch (state) {
+        case "granted":
           showLocalNewsWithGeolocation();
-        } else if (result.state == 'prompt') {
+          break;
+        case "prompt":
           showButtonToEnableLocalNews();
-        }
-        // Don't do anything if the permission was denied.
-      });
-    &lt;/script&gt;
+          break;
+        default:
+          // Don't do anything if the permission was denied.
+          break;
+      }
+    });
   </pre>
   </div>
   <div class="example" id="example-notifications">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -912,16 +912,19 @@ spec: webidl
...
-      navigator.permissions.query({name:'geolocation'}).then(function(result) {
-        if (result.state == 'granted') {
-          showLocalNewsWithGeolocation();
-        } else if (result.state == 'prompt') {
-          showButtonToEnableLocalNews();
-        }
-        // Don't do anything if the permission was denied.
-      });
-    &lt;/script&gt;
+navigator.permissions.query({ name: "geolocation" }).then(({ state }) => {
```
jyasskin: With bikeshed, you can leave things inside `<pre>` indented, and it'll figure out the indentation level.

index.bs
```diff
@@ -953,33 +954,42 @@ spec: webidl
...
+    console.log("Geolocation permission state is:", geoState);
+    console.log("Notifications permission state is:", notifState);
+  });
   </pre>
   </div>
   <div class="example" id="example-deviceId">
   <p>
-    This example is checking whether the page has the {{"camera"}}
-    permission for the first device in the list. Error checking omitted.
+    This example is checking the permission of the the first camera.
```
jyasskin: s/is checking/checks/? 

Can you say what "the permission of" means. Is it permission to record video from, permission to see the metadata of, or what?

You have a "the the".

index.bs
```diff
@@ -953,33 +954,42 @@ spec: webidl
...
-    &lt;/script&gt;
+navigator.mediaDevices
+  .enumerateDevices()
+  // filter out just video inputs
+  .then(
+    devices => devices.filter(({ kind }) => kind === "videoinput")
+  )
+  // get the deviceId of first camera, create query object
+  .then(
+    ([{ deviceId }]) => ({ name: "camera", deviceId })
```
jyasskin: It seems unnecessarily verbose to create the query object and call `query()` in two separate `.then()` steps. Combine them?

index.bs
```diff
@@ -953,33 +954,42 @@ spec: webidl
...
+  // perform query
+  .then(
+    queryObj => navigator.permissions.query(queryObj)
+  )
+  // log the state
+  .then(
+    ({ state }) => console.log("Permission state of first camera is", state)
+  )
+  .catch(
+    (err) => console.error(err.stack)
```
jyasskin: You don't need the ()s around `err` here. Do we want to be consistent about being completely minimal, or do we not care?

jyasskin:
Otherwise, looks good to me.

jyasskin:
Oh, except, what's "chore(...):" in the commit titles? If it's just personal bookkeeping, we should probably keep it out of the commit log.

index.bs
marcoscaceres: Will re-indent.

index.bs
marcoscaceres: Yep, good suggestion. Can do basically, filter, and then map to query object, and then map to promise there... and promise.all() for the cameras.

index.bs
marcoscaceres: Let's go with minimal... no convention has really emerged yet, AFAIK, regarding these shorthands... but omitting optional bits sounds good unless they are actually needed.

marcoscaceres:
> Oh, except, what's "chore(...):" in the commit titles? If it's just personal bookkeeping, we should probably keep it out of the commit log.

It's a convention for better tracking of changes, and for tracking the context of where changes are made. It's loosely based on [Karma's commit conventions](http://karma-runner.github.io/0.10/dev/git-commit-msg.html). I would encourage us to adopt it because it makes it easier to understand what has changed and why.

marcoscaceres:
... refactor(examples) is probably a more accurate description of what changed here.

marcoscaceres:
Addressed the feedback... please merge if all looks ok now.

Issue #85, opened 2016-04-21
Permission API has terminology issues one could drive a truck through.


jan-ivar:
Rather than rehash it here, I ask that people read what I just posted in https://github.com/w3c/mediacapture-main/issues/334#issuecomment-212999808, as context is important.

#### Ask:
 1. Rename this spec to "Persistent Permissions API".
 1. Define and use the term "persistent permission" instead of "permission" throughout.
 1. Clarify how "access" is distinct from "persistent permission".
 1. Avoid use of the ambiguous term "permission".
 1. Avoid conflating a *persistent permission* API with an *access* API.
 1. Remove attempts to co-opt existing APIs for `request` https://github.com/w3c/permissions/issues/83 or `revoke` https://github.com/w3c/permissions/issues/46 of ***access***.

jyasskin:
This API has context for calls to functions like `query()`, `request()`, and `revoke()`: the current [environment settings object](https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object). Saying that `query()` only refers to the persistent part of a permission like geolocation would break its use in libraries that could take advantage of the current tab having permission, even if there's no persistent permission. (There is, of course, the privacy issue #52, but we shouldn't split discussion of that.)

I believe the original intent of this spec was to deal with both temporary and persistent permissions. It doesn't do a great job of that yet, because of the description of a [single permission store](https://w3c.github.io/permissions/#permission-store), but I think that's fixable.

I do think we might want to rename this spec to just "Permissions": it defines the platform permission architecture in addition to an API, and other specs, like Media Capture, are likely to reference this one primarily to interact with that architecture, not because they want to use the API. It's a little like https://storage.spec.whatwg.org/ in that regard.

I'll leave this open for a day or two to let other people chime in, but I don't expect to take these suggestions.

jan-ivar:
Please see beyond my recommendation to the issues I point out. This is after all an *issue* not a *PR*, and I believe it points out real terminology issues with the spec:

 * There is no definition of *permission*.
 * The implicit definition is that of a *persistent permission*, which is fine, but should be made clear.
 * Statements from participants about this covering access ("temporary permission") don't hold.
 * Algorithms based on this poor definition, such as *request*, make no sense (see below).

We should leave it open until these are addressed. Again, my main explanation about the poor terminology is laid out in https://github.com/w3c/mediacapture-main/issues/334#issuecomment-212999808, not here, so please go read that first.

[...]

Issue #83, opened 2016-04-18
Drop .request()


martinthomson:
I realize that this might be a minority opinion here, but I think that `.request()` stands to undermine a lot of the work that we've done to make permissions on the web meaningful and relevant to users.  The most important aspect of that being that permissions requests are made in context.

You have probably all read [this](http://robert.ocallahan.org/2011/06/permissions-for-web-applications_30.html) before, but it's worth a refresher.  That summarizes where this is coming from.  I will note that model has been vindicated in that Android has moved to making requests in context also.

What I want to focus on here, something I consider more important than speculation about how applications might actually use `.request()`, is the consequence aspect of making individual APIs trigger use consent interactions.

When an application calls `getUserMedia()` (for the first time), it causes _two_ things to happen: a user consent interaction, and acquisition of a resource.  The same for geolocation.  As a consequence, the users sees that the application is then recording media or accessing their GPS (both of which usually have indicators in chrome that the user has already learned to recognize).

I believe that coupling of request and consequence to be important.  It provides applications with an incentive to defer requests until the point where the capability provided by the API is needed.  More so in browsers that don't implement persistent-by-default permissions.  That preserves the contextual relationship between asking and using.

An API like `.request()` undermines this model.  We already have ways to request permission from users, and I don't see any suggestion that we're going to change that.  That means that this API only serves to endorse modes where asking is decoupled from consequence.

There might be exceptions to this, and here I note that we don't appear to have a good substitute for just asking ahead of time when it comes to push notifications.  The Notifications API has its own analogue of `.request()` for that purpose, which turns out to be bad UX, but the best we've been able to come up with thus far.  That doesn't justify the creation of a generic permission request capability.

jimmywarting:
I like request method because how it unifies the API on how you request permissions. Read #48 as to why I hate the existing individual request permission model. Each one of them have there own syntax and some of them are "broken" or lacks useful informations as "was the permission rejected?"

jan-ivar:
> Asking ahead of time will be entirely up to the developers. I think most developers has a common sense and would only disturb the user with a prompt when needed

@jimmywarting why do common sense developers need `.request` then?

jyasskin:
[...]

PR #107, opened 2016-07-15
feat(prompts): allow UAs to deny non-gesture prompts

### LINKED ISSUE #77: Only allow permissions.request() during user interaction
Some quotes from Google:

> We believe users should only see a permission prompt if they’ve demonstrated willingness to grant that permission, or at minimum, intent to use a feature which requires that permission. In particular, we strongly advise against triggering permissions prompts on page load.
> 
> Showing users permission prompts they aren’t interested in is distracting and spammy. To ensure Chrome users get the best experience on the web we’re exploring automatically denying permission prompts from sites that display prompts which users largely ignore or deny (or worse, revoke).

from https://docs.google.com/document/d/1WNPIS_2F0eyDm5SS2E6LZ_75tk6XtBSnR1xNjWJ_DPE/edit#heading=h.55stxhevs7pi

> Always request access to location on a user gesture

from https://developers.google.com/web/fundamentals/native-hardware/user-location/user-consent

This new API seems like the perfect opportunity to enforce this restriction to avoid spammy popups: Permission requests shouldn't appear if `permissions.request()` isn't called as a result of user interaction. The older permission APIs can eventually be deprecated in favour of .request().

jyasskin:
Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/gesture-may-gate-prompt/index.bs#requesting-more-permission

This allows a UA to leave the permission state at "prompt" but deny
permission requests that weren't triggered from a user gesture without
prompting the user. We don't require UAs to deny these prompts because
it would cause non-persistent permissions to require two click in
some flows that Firefox folks think should only require one click.

Fixes #77.

@mnoorenberghe @noncombatant @jan-ivar @benfredwells @martinthomson

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/permissions/pull/107.html" title="Last updated on Feb 16, 2021, 11:20 PM UTC (819bd27)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/107/f9efa43...jyasskin:819bd27.html" title="Last updated on Feb 16, 2021, 11:20 PM UTC (819bd27)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -48,6 +48,7 @@ spec: html
         text: in parallel
         text: origin; for: /
         text: queue a task
+        text: triggered by user activation
 spec: ui-events
     type: dfn
         text: user agent
@@ -290,6 +291,10 @@ spec: webidl
           <var>current state</var> and abort these steps.
         </li>
         <li>
+          If this algorithm is not <a>triggered by user activation</a>,
+          the UA may return {{"denied"}} and abort these steps.
+        </li>
+        <li>
           Ask the user's permission for the calling algorithm to use the
           <a>powerful feature</a> described by |descriptor|.
         </li>
@@ -335,6 +340,10 @@ spec: webidl
           choose">prompts for the user to choose</a> from the same set of
           options with the same |descriptor| must return the same option, unless
           the UA receives <a>new information about the user's intent</a>.
+        </li>
+        <li>
+          If this algorithm is not <a>triggered by user activation</a>,
+          the UA may return {{"denied"}} and abort these steps.
         </li>
         <li>
           Ask the user to choose one of the options or deny permission, and wait


marcoscaceres:
don't we want consistency here across browsers? The "may" gives no certainty.

marcoscaceres:
That question above was, I guess, for "Firefox folks".

jan-ivar:
Are we presupposing here that UAs are barred from denying non-gesture prompts today? That seems false, because a user agent by definition can act on behalf of its user.

Permission prompts are the browser’s own communication with its user, and dictating such browser-chrome UX is traditionally not the domain of specs.

This spec started as a web surface, but seems to want to inflict change outside of that scope.

marcoscaceres:
> Are we presupposing here that UAs are barred from denying non-gesture prompts today? That seems false, because a user agent by definition can act on behalf of its user.

No, the question is about the behavior of calls to `.request()` to get more permission, and if calls should only succeed if initiated by user action. 

> Permission prompts are the browser’s own communication with its user, and dictating such browser-chrome UX is traditionally not the domain of specs.

Yes, but they are initiated by scripts (e.g., `geolocation.getCurrentPosition()`).

> This spec started as a web surface, but seems to want to inflict change outside of that scope.

The spec doesn't "want" anything - the people editing the spec are just asking for input. You are making it sound like there is malicious intent in what is currently in the spec (which is not the case!). We are just looking for input to see if `.request()` is workable and worth keeping.

annevk:
@jan-ivar I think what we want is something that is good for users and predictable for developers. The latter requires some consistency across browsers. That's why folks are trying to figure out the constraints.

jan-ivar:
@marcoscaceres My understanding is that the ***request permission to use*** algorithm that is modified here, is for use by [other specs](https://w3c.github.io/mediacapture-main/getusermedia.html#request-permission) and not limited to `.request()`, hence my concern about scope.

jan-ivar:
@annevk I find the application of gesture here misplaced and poorly thought out.

I understand we sometimes put a *feature* behind a *gate*, to preserve user intention.

User gesture is a *gate*.
Fullscreen is a *feature*.
Camera access is a *feature*.
Permission prompt is a *gate*.

Putting a gate behind a gate, is unprecedented and a bad idea, and pushes users to leave the gate open.

annevk:
I think there's a reasonable case for both sides of the argument.

* Showing users a dialog without the user having asked for it might be confusing.
* If the user already expects a dialog (e.g., because of the domain or nature of task at hand), then having to interact with the page and then the dialog is convoluted.

What's the case for requiring user interaction with the page first?

jan-ivar:
@annevk please see https://github.com/w3c/permissions/issues/77 so we don't discuss this twice (it mentions at least [one use-case](https://github.com/w3c/permissions/issues/77#issuecomment-223095222)).

This was also discussed on an email thread with permission and UX people at Mozilla, where we found that deep linking (from bookmarks, links from chats and other sites) is a valid use-case for repeat visitors, which is made worse only for (most) Firefox users (since unlike Chrome, Firefox supports non-persistent grants, and offers it by default).

Importantly, it's not the *feature* that Google wants to restrict to a user gesture, only the (to them initial) permission prompt, so you won't hear them argue there are no use-cases.

In fact, I'm sure Google will continue to support e.g. geo-location on page-load for all repeat visitors (read: implicit persistent permission granters). The only losers are Firefox users who resist granting sites persistent access to their whereabouts (still most Firefox users, given it's default).

Will this help prevent permission spamming? In Chrome, yes (marginally, since the remedy is half-baked in that it likely only delays an inevitable prompt), because it detects first-time visits really well.

In Firefox, this dragnet fails to distinguish first-time visitors from privacy-minded visitors, so it hurts a segment that had nothing to do with spam.

jan-ivar:
@annevk apologies if I got your question backwards, and it's for someone else.

annevk:
The question was mainly meant for Chrome and like-minded browsers, but the reminder pointer to #77 helped.

The problem I see with the proposal from Google to "allow different models" is that sites will likely code toward a particular model and since Chrome has way more outreach the Firefox approach likely loses over time, which seems detrimental to users who don't want to grant things persistently.

jyasskin:
> Are we presupposing here that UAs are barred from denying non-gesture prompts today? That seems false, because a user agent by definition can act on behalf of its user.

Say you call query('foo'), then you call requestFoo() without a gesture, and the UA decides to auto-deny, then query('foo') again, then requestFoo() again with a gesture, and the UA does show a prompt. What [permission state](https://w3c.github.io/permissions/#permission-state) did the two query() calls return?

jan-ivar:
@jyasskin you tell me. Does `requestFoo()` mean "request feature X" or "request permission elevation for feature X" ? Those two seem to be conflated a lot. See also https://github.com/w3c/permissions/issues/83.

jyasskin:
`requestFoo()` could be `navigator.geolocation.getCurrentPosition()`, for example.

jyasskin:
For that matter, the difference between "request permission to use 'foo'" and "request that the permission state for 'foo' be increased" doesn't affect the answer here. This is about what happens before the user chooses, while that distinction is about after.

jan-ivar:
Thanks for clarifying, though I'm still not sure this question is for me, or how its answer bears on this conversation, sorry.

jyasskin:
@jan-ivar In https://github.com/w3c/permissions/pull/107#issuecomment-234170211 you asked "Are we presupposing here that UAs are barred from denying non-gesture prompts today?" The answer is "yes, UAs are barred from denying non-gesture prompts today" because there's no set of permissions states allowed by the current spec that would allow the sequence of events in https://github.com/w3c/permissions/pull/107#issuecomment-234320049.

"inflict change" is a pejorative way to put it, but I'm trying to specify how browsers handle permissions in general, outside the scope of the `navigator.permissions` object (contra @marcoscaceres' answer). I think this PR makes the spec more closely align with and describe the full variance in what browsers are actually doing.

@annevk I think you [captured](https://github.com/w3c/permissions/pull/107#issuecomment-234238254) one of the reasons to require user interaction with the page first: "Showing users a dialog without the user having asked for it might be confusing." Other reasons are behind the links in https://github.com/w3c/permissions/issues/77#issue-147599096.

jan-ivar:
@jyasskin Sorry, I meant "today" as in [this spec having no consensus](https://github.com/w3c/permissions/issues/85#issuecomment-213616988) today.

"Trying to specify how browsers handle permissions in general" is perhaps an approbative way to put that we're locking down how all future browsers are allowed to manage permissions on behalf of their users - for all features - based on how browsers manage this today. That's a change.

Is that the intention of this spec? If so, it could be made clearer, as there seems to be some confusion about what we're actually discussing among participants.

jyasskin:
We're in the process of [publishing the editors' draft to https://www.w3.org/TR/permissions/](https://lists.w3.org/Archives/Public/public-webappsec/2016Aug/0041.html), and it'd be nice to get this relaxation in either before or soon after that. Are there any remaining objections to this PR?

annevk:
I still think https://github.com/w3c/permissions/pull/107#issuecomment-234256944 is problematic. By allowing this for certain user agents, there's a huge risk that other user agents will have to change their permission model to match, to the detriment of their users.

It also ties back to trying to figure out a generic "request access" primitive, if there is such a thing. E.g., for persistent storage it's likely Chrome might never even show a dialog. So at that point you can't even rely on a generic model around `request()` for a single browser, since it'll vary based on the arguments you pass. That seems really icky.

jyasskin:
@annevk I think the risk you're worried about goes like:

1. Chrome starts enforcing a gesture.
2. Developers notice that when they test on Firefox, their page doesn't work on Chrome.
3. Firefox is forced to add a gesture-enforcing mode lest developers all switch to Chrome.
4. Websites change their UI so that permissions are always requested with a gesture, so nobody's actually taking advantage of Firefox's flexibility.
5. People complain about the double-prompt in Firefox, and the number of people who select "remember this grant" in Firefox increases.
6. Firefox's UI folks give in and make "remember this grant" the default.

Is that right?

I suspect Chrome's security team is going to insist on the gesture even if the spec disallows it, but we could wait to make the spec allow it until that's a change to reflect reality.

PR #109, opened 2016-07-18
feat(registry): features require a secure context by default, with opt-out

jyasskin:
By default, features will always be "denied" in non-secure contexts. The
"allowed in non-secure contexts" flag undoes this default.

It would be possible to throw an exception from request() and possibly
query() when a secure-only feature is used from a non-secure context,
but denying leads to one fewer behavior that developers need to account
for.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/secure-permissions/index.bs#requires-a-secure-context

@mikewest

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -213,9 +213,13 @@ spec: webidl
       algorithm should succeed without prompting the user, show the user a
       prompt to decide whether to succeed, or fail without prompting the user.
       The UA must return whichever of these values most accurately reflects the
-      user's intent. Subsequent uses of |descriptor|'s <a>permission state</a>
-      with the same <a>current settings object</a> must return the same value,
-      unless the UA receives <a>new information about the user's intent</a>.
+      user's intent, except that if the <a>current settings object</a> is a
+      <a>non-secure context</a> and
+      <code>|descriptor|.{{PermissionDescriptor/name}}</code> isn't <a>allowed
+      in non-secure contexts</a>, then the UA must return {{"denied"}}.
+      Subsequent uses of |descriptor|'s <a>permission state</a> with the same
+      <a>current settings object</a> must return the same value, unless the UA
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -350,9 +354,15 @@ spec: webidl
...
     Each enumeration value in the {{PermissionName}} enum identifies a
     <a>powerful feature</a>. Each <a>powerful feature</a> has the following
-    permission-related algorithms and types:
+    permission-related flags, algorithms, and types:
   </p>
   <dl>
+    <dt>A <dfn export>requires a secure context</dfn> flag</dt>
+    <dd>
+      Indicates that the feature will always be {{"denied"}} in <a>non-secure
+      contexts</a>. This flag defaults to un-set.
```
jyasskin: Arguably, we should default this to "set", and make features explicitly say that they don't require a secure context... Then I'd probably reword it to reverse the sense: "allowed in non-secure contexts" defaulting again to un-set.

Thoughts?

martinthomson: I would prefer to default to disabling these features in non-secure contexts.

jyasskin: Done, thanks.

index.bs
```diff
@@ -517,7 +530,8 @@ spec: webidl
       The <dfn>"camera"</dfn>, <dfn>"microphone"</dfn> , and
       <dfn>"speaker"</dfn>
       permissions are associated with permission to use media devices as
-      specified in [[GETUSERMEDIA]] and [[audio-output]].
+      specified in [[GETUSERMEDIA]] and [[audio-output]]. {{"speaker"}} is
```
jyasskin: @alvestrand Did I get the right set of media-related permissions that require secure contexts?

PR #115, opened 2016-07-27
feat(algorithms): allow calling with a PermissionName

jyasskin:
Instead of forcing specs for simple permissions to write out a permission
descriptor. This partially fixes @annevk's objection that writing `{name: "persistent-storage"}` in a spec doesn't formally create a WebIDL dictionary instance, although we'll still need to formalize syntax like that for more complicated descriptors.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/permission-name-shorthand/index.bs#reading-current-states. I'm not sure the best place to define this. I want folks who click the link to, say, https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/permission-name-shorthand/index.bs#request-permission-to-use to find the definitions of the shorthands (so a separate "shorthands" section is out), and I don't want to make every spec write out `<a for="PermissionName">request permission to use</a>` to disambiguate the two definitions (so direct links to the shorthand are out). The wording doesn't fit well right next to the main definition, so I put it below.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -242,6 +242,12 @@ spec: webidl
         href="https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects">several
         possible settings objects</a> it uses.
       </p>
+
+      <p>
+        As a shorthand, a {{PermissionName}} |name|'s <a>permission state</a> is
+        the <a>permission state</a> of a {{PermissionDescriptor}} with its
+        {{PermissionDescriptor/name}} member set to |name|.
+      </p>
     </div>
 
     <div algorithm>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
LGTM, but maybe we wait for @annevk to give an ok before merging.

annevk:
This looks okay, but doesn't really resolve the problem. It just hides it a bit more. The fact that you're using ECMAScript and IDL concepts in a place where ECMAScript and IDL are not processed (or defined to be processed) is what is problematic.

jyasskin:
"a `PermissionDescriptor` with its `name` member set to *name*." seems no worse than any other time we create a WebIDL thing from prose, and is fairly close to the wording in http://heycam.github.io/webidl/#es-dictionary. Or have I missed a subtlety?

I agree that this change doesn't fix the problem of Web Bluetooth creating a descriptor in step 3 of [request Bluetooth devices](https://webbluetoothcg.github.io/web-bluetooth/#request-bluetooth-devices). I've filed https://github.com/heycam/webidl/issues/142 to try to get a good syntax for that use.

annevk:
Ah okay, if you drop the syntax it is a lot better, yes.

PR #118, opened 2016-08-03
refactor: move permissions registry (closes #103)

### LINKED ISSUE #103: Move the Permissions registry
The Permissions Registry being in the middle of the spec is quite distracting. It adds a lot of unnecessary noise right in the middle of the spec, before the interfaces have been introduced, etc.... making the spec more difficult to follow than it should be.

Can I suggest that we move it to the end of the document or remove the registry entirely? IMHO, each registry item should be moved (normatively) to their respective specs. We could then link to those specs, non-normatively, but the registry doesn't really belong in this spec.

marcoscaceres:
Move Permissions Registry just before the Examples section.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -368,6 +368,261 @@ spec: webidl
 </section>
 
 <section>
+  <h2 id="status-of-a-permission">
+    Status of a permission
+  </h2>
+  <pre class='idl'>
+    enum PermissionState {
+      "granted",
+      "denied",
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Preview at: https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/w3c/permissions/move_perms/index.bs

PR #120, opened 2016-08-10
Resolve TODOs on media type permissions

alvestrand:
This likely needs some review before merging. I added what I thought was the obvious thing to say, and removed some overtaken-by-events materials, but there are enough things linked that I might have some of them wrong.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -895,46 +895,48 @@ spec: webidl
       </dt>
       <dd>
         A list of {{MediaDeviceInfo/deviceId}} values for the devices the user
-        has granted access to.
-      </dd>
-      <dt>
-        <a>permission result type</a>
-      </dt>
-      <dd>
-        TODO
+        has made a non-default decision on access to.
       </dd>
       <dt>
         <a>permission query algorithm</a>
       </dt>
       <dd>
-        TODO
+        The permission query algorithm runs the following steps:
+        <ol class="algorithm">
+        <li>
+        If |permissionDesc|.deviceId exists in the <a>extra permission data</a>,
+        set <code><var>status</var>.state</code>
+        to |permissionDesc|'s <a>permission state</a> and terminate these
+        steps.
+        </li>
+        <li>Set <var>global</var> to the {{PermissionDesc}} for the descriptor
+        that results from removing the "deviceId" value.
+        </li>
+        <li>
+        Set <code><var>status</var>.state</code> to <var>global</var>'s
+        <a>permission state</a>.
+        </li>
       </dd>
       <dt>
         <a>permission request algorithm</a>
       </dt>
       <dd>
-        TODO
+        Run the <a>boolean permission request algorithm</a>
       </dd>
       <dt>
         <a>permission revocation algorithm</a>
       </dt>
       <dd>
-        TODO: Stop playing/recording data?
+        When permission for a device is revoked, the UA MUST run the
+        algorithm to stop the track for any other reason than the stop()
+        method being invoked for each MediaStreamTrack sourced from that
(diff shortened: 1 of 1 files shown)


jyasskin:
Thanks! I'd prefer for this to live in the getUserMedia spec (like I'm doing in https://github.com/whatwg/storage/pull/36) so that you can update it without getting the permission spec editors' permission, and so the media experts naturally find it to review it. But I'm happy to review this PR and commit it here if it doesn't fit there quite yet.

index.bs
```diff
@@ -895,47 +895,49 @@ spec: webidl
...
+        If |permissionDesc|.deviceId exists in the <a>extra permission data</a>,
+        set <code><var>status</var>.state</code>
+        to |permissionDesc|'s <a>permission state</a> and terminate these
+        steps.
+        </li>
+        <li>Set <var>global</var> to the {{PermissionDesc}} for the descriptor
+        that results from removing the "deviceId" value.
+        </li>
+        <li>
+        Set <code><var>status</var>.state</code> to <var>global</var>'s
```
jyasskin: Do you want to have a way for the site to request the list of all granted `deviceId`s? If so, you'll need to extend the permission result type too, with a list of deviceIds and/or MediaStreams.

index.bs
```diff
@@ -895,47 +895,49 @@ spec: webidl
...
+        <a>permission state</a>.
+        </li>
       </dd>
       <dt>
         <a>permission request algorithm</a>
       </dt>
       <dd>
-        TODO
+        Run the <a>boolean permission request algorithm</a>
```
jyasskin: I suspect `request()` isn't very useful when it can only request a device ID the site already knows about. Folks are likely to want to pass the same arguments they can pass to `getUserMedia()`, and to get a stream back.

index.bs
```diff
@@ -895,47 +895,49 @@ spec: webidl
...
       </dd>
       <dt>
         <a>permission revocation algorithm</a>
       </dt>
       <dd>
-        TODO: Stop playing/recording data?
+        When permission for a device is revoked, the UA MUST run the
+        algorithm to stop the track for any other reason than the stop()
```
jyasskin: I don't quite follow this sentence. Can you link to the algorithm you mean? Is it https://w3c.github.io/mediacapture-main/#dom-mediastreamtrack-stop?

jyasskin: Oh, it's the algorithm just after https://w3c.github.io/mediacapture-main/#track-ended, which is what your `<!--TODO-->` means.

index.bs
alvestrand: I don't see a need for an application to get that list, no. If it really wants one, it can call enumerateDevices and then do query() for each of them. I want to keep the API as simple as possible.

index.bs
alvestrand: Device IDs are already exposed through mediaDevices.enumerateDevices. I don't want to have to expose the complexity of constraints to the permission handling; the getUserMedia algorithm can decide which devices are potential candidates before asking the user to choose between them.

index.bs
alvestrand: Right. I think we should expose a name for it, so that it's easy to link to. (We may already have the anchor.)

index.bs
jyasskin: Ok.

index.bs
jyasskin: Ok.

index.bs
jyasskin: Let's wait until you have that link target, and then I'll merge this. Thanks!

jyasskin: I'm giving up on the link target for now, and just merging this, since it's an improvement.

PR #122, opened 2016-08-26
refactor(media devices): use "allowed to use"

### LINKED ISSUE #121: Use HTML's new "allowed to use" algorithm
It was introduced with https://github.com/whatwg/html/pull/1492 and would replace the bit about the `allowusermedia` attribute in https://w3c.github.io/permissions/#media-devices

See https://github.com/whatwg/fullscreen/issues/41 for example wording, once there's a PR.

jyasskin:
Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/allowed-to-use/index.bs#media-devices.

@foolip, have you considered how "allowed to use" interacts with worker contexts? 

@alvestrand, this touches the spec close to #120 which is still waiting on a link target in [mediacapture-main]. I think it's compatible, but FYI.

Fixes #121

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -30,9 +30,6 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
         text: current realm; url: current-realm
     type: interface
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
-spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
-    type: dfn
-        text: ancestor browsing context
 spec: mediacapture-main; urlPrefix: https://w3c.github.io/mediacapture-main/#
     type: attribute
         for: MediaDeviceInfo; text: deviceId
@@ -40,9 +37,11 @@ spec: mediacapture-main; urlPrefix: https://w3c.github.io/mediacapture-main/#
         for: MediaDevices; text: getUserMedia(); url: dom-mediadevices-getusermedia
 </pre>
 <pre class="link-defaults">
+spec: dom
+    type: interface
+        text: Document
 spec: html
     type: dfn
-        text: browsing context container
         text: event handler
         text: event handler event type
         text: in parallel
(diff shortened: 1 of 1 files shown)


foolip:
No, I hadn't thought about that at all. Do you think there features that will run into this? CC @annevk

jyasskin:
@foolip Yeah, I think service workers will eventually be able to get permission to use features that are also enabled via iframe attributes. The approach I took in this patch, allow use if the current global doesn't have an associated `Document` will probably work in most cases, and we may have to update it again to deal with Feature-Policy, so I don't think a fix to HTML is urgent.

annevk:
That approach would mean that a service worker for an iframe that does not have access would get access. Seems bad?

foolip:
I'm a bit confused. The only two features controlled by "allowed to use" are Fullscreen and getUserMedia, and those are only available in document contexts. Fullscreen can't make sense elsewhere, but getUserMedia could, is that the case under consideration?

raymeskhoury:
I think it's probably better to assume service workers can't use features that have permission for the moment and figure them out later. I think there are too many caveats that we haven't though through.

> but getUserMedia could, is that the case under consideration?

Not that I'm aware of.

jyasskin:
Yeah, nothing's planned concretely yet. We can definitely wait to figure it out.

@annevk There are potentially two ways of granting permission to service workers: 1) any time they're running, even if they don't have a foreground tab open, or 2) when they have a foreground window. Checking for a foreground window would naturally use "allowed to use" on that window, which would exclude windows that are disallowed iframes.

jyasskin:
I'm waiting on an LGTM for this change, which I think doesn't rely on figuring anything out about permissions in service workers.

marcoscaceres:
LGTM

alvestrand:
This requires major archeaology in definition chains in order to figure out whether it's a chain or not. I trust that it's legitimate. LGTM.

PR #125, opened 2016-08-30
Pass objects instead of PermissionDescriptors to the Permissions interface.

### LINKED ISSUE #59: Passing base dictionary wont work.
Same thing as happened in `getStats`, a WebIDL method taking a base `PermissionDescriptor` dictionary wont see any members from derived dictionaries, e.g. `deviceId` from `DevicePermissionDescriptor`, so this API needs to take `object`, interfaces, or be rethought.

Also https://github.com/w3c/permissions/issues/42 may be related.

jyasskin:
Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/permissions/object-params/index.bs#permissions-interface.

@tobie, since this is an attempt to bring us into compliance with WebIDL, could you take a look? Is there a better way to call the right part of the "converted to an IDL value" algorithm?

Fixes #59.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -477,11 +477,11 @@ spec: webidl
   <pre class='idl'>
     [Exposed=(Window,Worker)]
     interface Permissions {
-      Promise&lt;PermissionStatus&gt; query(PermissionDescriptor permissionDesc);
+      Promise&lt;PermissionStatus&gt; query(object permissionDesc);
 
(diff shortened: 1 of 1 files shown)


tobie:
This LGTM, but I'm really not an authority here (yet). Think @annevk worked on similar APIs for the URL spec and might be of greater assistance here.

PR #138, opened 2017-03-03
Add permission for Ambient Light Sensor.

tobie:


anssiko:
@jyasskin this is a gentle nudge to get this landed, is blocking https://github.com/w3c/ambient-light/pull/21

raymeskhoury:
Do we want 1 permission for all sensors in the API or one for each sensor type? I actually don't have a strong feeling right now.
@benfredwells

raymeskhoury:
I chatted with @benfredwells and I think having a separate one per sensor type is probably a better approach given the different security profile of each.

PR #143, opened 2017-04-26
chore(perm interface): Fix return value handling

jugglinmike:
Unlike the other methods defined in this section, `revoke` is defined as
a series of steps to be run in parallel. Returning a value is
meaningless in this context. Instead, the result of the operation should
be communicated via the Promise instance that is defined and
synchronously returned prior to the parallelized steps.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -580,14 +580,14 @@ spec: webidl
   <ol>
     <li>Let |rootDesc| be the object |permissionDesc| refers to, <a>converted to
     an IDL value</a> of type {{PermissionDescriptor}}. If this throws an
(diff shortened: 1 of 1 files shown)


jyasskin:
Sorry it took me so long to look at this. The "real" problem here is that `revoke` shouldn't run all of its steps in parallel. We're specifically not allowed to use "converted to an IDL value" from parallel steps, since ECMAScript objects are single-threaded.

It should go parallel before step 3, and then have step 5 call into the middle of query() with the already-converted *typedDescriptor*.

I'm finally working on pulling `request()` and `revoke()` into an incubation, and I'll fix this there.

PR #155, opened 2017-09-13
Add ClipboardPermissionDescriptor

garykac:
***
[Preview](https://s3.amazonaws.com/pr-preview/garykac/permissions/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/permissions/025d298...garykac:9098b5a.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1089,18 +1089,18 @@ spec: webidl
       <dd>
         <pre class='idl'>
           dictionary ClipboardPermissionDescriptor : PermissionDescriptor {
-            boolean fullAccess = false;
+            DOMString access = "write";
           };
         </pre>
         <p>
-          `{name: "clipboard", fullAccess: true}` is <a>stronger than</a>
-          `{name: "clipboard", fullAccess: false}`.
+          `{name: "clipboard", access: "full"}` is <a>stronger than</a>
+          `{name: "clipboard", access: "write"}`.
         </p>
         <p>
           The permission for basic (write only) clipboard access is
-          `{name: "clipboard"}`.
+          `{name: "clipboard", access: "write"}`.
           The permission for full (read/write) clipboard access is
-          `{name: "clipboard", fullAccess: true}`.
+          `{name: "clipboard", access: "full"}`.
         </p>
       </dd>
     </dl>
--- a/index.bs
+++ b/index.bs
@@ -1107,6 +1107,12 @@ spec: webidl
           The permission for full (read/write) clipboard access is
           `{name: "clipboard", access: "full"}`.
         </p>
+        <p>
+          See the
+          <a href="https://w3c.github.io/clipboard-apis/#clipboard-permissions">Clipboard Permissions</a>
+          section of [[clipboard-apis]] for additional information on how this permission
+          is used.
+        </p>
       </dd>
     </dl>
   </section>
--- a/index.bs
+++ b/index.bs
@@ -1090,11 +1090,12 @@ spec: webidl
         <pre class='idl'>
           enum ClipboardPermissionAccess {
             "full",
+            "read",
             "write",
           };
           
           dictionary ClipboardPermissionDescriptor : PermissionDescriptor {
-            ClipboardPermissionAccess access = "write";
+            ClipboardPermissionAccess access;
           };
         </pre>
         <p>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1082,6 +1082,28 @@ spec: webidl
...
       asynchronous methods in the [[clipboard-apis]].
     </p>
+    <dl>
+      <dt>
+        <a>permission descriptor type</a>
+      </dt>
+      <dd>
+        <pre class='idl'>
+          dictionary ClipboardPermissionDescriptor : PermissionDescriptor {
+            boolean fullAccess = false;
```
marcoscaceres: Are you sure that in the future you won't have a requirement for more granular access? Did you consider instead:

```JS
{
   access: "full",
}
```

garykac: We don't believe that we'll need to be more granular than write (basic) and read/write (full), partially because that would be hard to surface to the user in a clear way.

Having said that, if you think it would be better to have this be `{ access: "write" }` and `{ access: "full" }` (or "read-write") then I don't have a problem with that.  The default `{ clipboard }` would be equivalent to `{ name: "clipboard", access: "write" }` and `full` would be stronger than `write`.

In the future, we may want to add another dimension (e.g., to control access to particular mimetypes), but that would be in a separate field.

marcoscaceres: Thanks for the clarification - that makes sense. I don't have a strong opinion, so I think it's probably ok to just have `fullAccess` to stand for "read-write" (and missing for just write). 

Having said that, this is the first permission to require this kind of read/write access pattern, so I want to make sure we end up with something that future specs can also use. 

@jyasskin, would appreciate your input here...

garykac: Agreed, and `{ access: "full" }` does seem to be a bit more future-proof if we ever need to make changes.

Also, see https://github.com/w3c/clipboard-apis/issues/51 for additional context about why we don't think separate "read" and "write" works well. For some reason, I thought you were on that thread.

index.bs
```diff
@@ -1082,6 +1082,39 @@ spec: webidl
...
+      </dt>
+      <dd>
+        <pre class='idl'>
+          enum ClipboardPermissionAccess {
+            "full",
+            "write",
+          };
+          
+          dictionary ClipboardPermissionDescriptor : PermissionDescriptor {
+            ClipboardPermissionAccess access;
```
marcoscaceres: I wonder if this should have the default ` ClipboardPermissionAccess access = "write"`? It takes the ambiguity out of what it means for it to be missing (which implies "write").

garykac:
I pinged @jyasskin earlier today and he said he was out until Tuesday.

jyasskin:
FWIW, I'd rather put the Descriptor specifications in their own specs, so in https://w3c.github.io/clipboard-apis/ for this one. That lets the spec authors keep everything up to date without blocking on the permissions editors. You can imitate https://webbluetoothcg.github.io/web-bluetooth/#permission-api-integration.

This spec should link to the Clipboard definition in the PermissionName enum, but that's currently broken for `"bluetooth"` and `"persistent-storage"`. I'll see if I can fix that.

martinthomson:
See comments on the [clipboard API issue](https://github.com/w3c/clipboard-apis/issues/51#issuecomment-331301019).

marcoscaceres:
Can someone give me a tl;dr of what the next steps are here? Is this to be merged or closed?

raymeskhoury:
We're trying to iron out some related questions in https://github.com/w3c/clipboard-apis/issues/51#issuecomment-329090035

saschanaz:
Should this be updated to match [this change](https://github.com/w3c/clipboard-apis/commit/aa1f8ac45ffc3ff9035fa64519c5c822ab28469c)?

dway123:
+1 to saschanaz. Since an updated permission has been [merged](https://github.com/w3c/clipboard-apis/commit/aa1f8ac45ffc3ff9035fa64519c5c822ab28469c) after the last activity on this PR's diff, and this PR has been inactive for several years, are we still planning on pursuing this change? If so, can we please update this change? And if not, can we please delete this change? Thanks!

marcoscaceres:
This should be updated only if as far as there is not duplicate text between the specs... and this looks quite a lot like:
https://www.w3.org/TR/clipboard-apis/#clipboard-permissions

Should we maybe just close this?

marcoscaceres:
@dway123, you probably now know best what's in both specs... let me know how you want to proceed - or if you need to take over this PR too?

saschanaz:
I think #209 was sufficient.

marcoscaceres:
Ok, thanks @saschanaz. Closing this... @dway123, let us know if you disagree.

dway123:
Thanks, @marcoscaceres. Closing this sounds good. 

As an aside, I think this would have been something like the nice single "clipboard" permission with a descriptor for "read" and "write" that we could have had as an alternative for "clipboard-read" and "clipboard-write", but it's indeed no longer relevant as "clipboard-read" and "clipboard-write" have already been merged now.

PR #159, opened 2017-09-22
Remove permissions.request() and permissions.revoke().

### LINKED ISSUE #46: Consider removing Permissions.revoke().
The introduction says:

> The initial intent of this document was to allow web applications to request and revoke permissions explicitly in addition to querying the permission status. This is an aspect of the specification that was controversial thus removed from the current document in a spirit of incremental changes: settling on a small API that can be improved. 

However, these exist in the document.  Should they be removed?  (I would say "yes".)

### LINKED ISSUE #83: Drop .request()

jyasskin:
They've moved to incubations in https://github.com/jyasskin/permissions-request
and https://github.com/jyasskin/permissions-revoke, which I'll post to WICG once this goes in.

This fixes #46 and fixes #83.

Sorry it took me so long to get around to this. @martinthomson @jan-ivar, did I catch everything you don't want in this spec? @slightlyoff @raymeskhoury @mounirlamouri, did I include everything in the two incubations that you want to see there? (And that was written down at all; they're clearly missing a bunch of examples.)

***
[Preview](https://s3.amazonaws.com/pr-preview/jyasskin/permissions/remove-request-revoke.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/permissions/12c9d6a...jyasskin:8d9ae4d.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -77,24 +77,28 @@ spec: webidl
   </h2>
   <p><em>This section is non-normative.</em></p>
   <p>
-    This document specifies a model and an API to query and request permissions
-    to powerful features on the Web platform. Web APIs have different ways to
-    deal with permissions. The [[notifications]] API allows developers to
-    request a permission and check the permission status explicitly.
-    Others expose the status to web pages when they try to use the API,
-    like the [[geolocation-API]] which fails if the permission was not
-    granted without allowing the developer to check beforehand.
+    This document specifies a model for permissions to use powerful features on
+    the Web platform and an API to query the current permission state of those
+    features.
   </p>
   <p>
-    The Permissions API provides tools for developers to control when permission
-    prompts are shown and to relinquish permissions that are no longer needed.
+    Current Web APIs have different ways to deal with permissions. For example,
+    the [[notifications]] API allows developers to request a permission and
+    check the permission status explicitly. Others expose the status to web
+    pages when they try to use the API, like the [[geolocation-API]] which fails
+    if the permission was not granted without allowing the developer to check
+    beforehand.
+  </p>
+  <p>
+    The {{Permissions/query()}} function provides a tool for developers to
+    control when permission prompts are shown.
   </p>
   <p>
     The solution described in this document is meant to be extensible, but isn't
-    meant to be applicable to all the current and future permissions
-    available in the web platform. Working Groups that are creating specifications
-    whose permission model doesn't fit in the model described in this document
-    should contact the editors by
+    expected to be applicable to all the current and future permissions
+    available in the web platform. Working Groups that are creating
+    specifications whose permission model doesn't fit in the model described in
+    this document should contact the editors by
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -781,35 +681,19 @@ spec: webidl
...
       existing instance of the <a>permission result type</a>, and updates the
       <a>permission result type</a> instance with the query result. Used by
-      {{Permissions}}' {{Permissions/query()}} method and the <a
+      {{Permissions}}' {{Permissions/query(permissionDesc)}} method and the <a
       href="#PermissionStatus-update">PermissionStatus update steps</a>. If
       unspecified, this defaults to the <a>boolean permission query
       algorithm</a>.
     </dd>
     <dt>
-      A <dfn export>permission request algorithm</dfn>
```
jan-ivar: We still need this algorithm I think. It's referenced in [mediacapture](https://w3c.github.io/mediacapture-main/getusermedia.html#request-permission).

jyasskin: I believe [mediacapture](https://w3c.github.io/mediacapture-main/getusermedia.html#request-permission) intends either https://w3c.github.io/permissions/#request-permission-to-use or https://w3c.github.io/permissions/#prompt-the-user-to-choose. @alvestrand?

jan-ivar: Right! It's the former. I'll open an issue.

PR #166, opened 2017-12-22
Discuss how query() makes abuse harder to detect.

### LINKED ISSUE #52: Privacy concern with being able to discern that a permission is granted
Nick Doty <npdoty@w3.org> to public-webappsec, public-privacy

Hi WebAppSec/Permissions folks, (CC Privacy Interest Group for their awareness and in case they have feedback)

I remain concerned about the privacy implications of this API in terms of checking for granted status. As I see it, the ability to confirm ahead of time that a request for user information won't provide a visible prompt to the user is likely to dramatically increase the frequency with which user information (for example, precise current geolocation) is requested without the user's awareness. We should have a good explanation for how to mitigate this issue before suggesting changes to widely-deployed implementations like Geolocation.

A reminder of the threat: a third-party script embedded into a page would like to know the user's precise geolocation, for example, for advertising or surveillance purposes, without the cooperation of the site owner. However, if the script calls navigator.geolocation.getCurrentPosition(), users of that site may be prompted, at which point the site owner is likely to be notified and the script is detected, altered or removed. For user agents that implement the Permissions API, an embedded script can query the permission status and only make the call if the status is granted; alternatively, it can register for notifications of any permission change, so that if the user ever does grant geolocation permission to that page (perhaps to see nearby info on a map), it can then access the user's geolocation. The user is not aware that location is being accessed by multiple scripts; the site owner is unlikely to be informed that an embedded script is accessing location.

This isn't a hypothetical situation. Last year when an advertising script on a prominent newspaper website was changed to request geolocation (and pass on lat/lon to the advertising provider), as a casual visitor I was able to see the change immediately, report it to the site owners and the change was rolled back within minutes.

Subscription to notifications of permission changes also has the potential for decreasing the effectiveness of ambient notices for detecting unexpected information access. If a user sees a compass or crosshair icon appear in their browser or OS at a time when not using a site's map interface, they might be surprised and inspect or revoke previously provided permissions. If embedded scripts can subscribe to notifications and only request location information at the exact moment that the site itself has requested geolocation information, an ambient notice won't indicate unusual activity to the user. This would be advantageous for an attacker injecting a surveillance script into a page, for example, who wants to evade detection by the user.

Providing access to "denied" permission status doesn't seem to incur this problem and could help developers who want to provide a more detailed explanation when a user has denied a permission. That would also be more useful for the case of revoking permission, which is an action that sites currently don't have any insight into. To remember when a permission has already been granted or to remember the different presentations to the user explaining the use of permissions, developers can use cookies or localStorage, as they do now to remember other preferences.

Thanks,
Nick

jyasskin:
Fixes #52.

@npdoty, does this handle your concern? Chrome and Firefox have [shipped](https://caniuse.com/#feat=permissions-api) the `query()` function, so it seems like the only thing to do is to suggest making use of permissions visible, rather than change the behavior of `query()`.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -124,6 +124,19 @@ spec: webidl
   </h2>
   <p><em>This section is non-normative.</em></p>
   <p>
+    Web pages often run more- and less-trusted components as the same origin.
+    For example, a newspaper may run advertising code without sandboxing it into
+    a cross-origin iframe. If the newspaper has a legitimate reason to use a
+    person's location, that also happens to grant access to the less trusted
+    advertiser. Without the {{Permissions/query()}} function in this
+    specification, to read the person's location, an advertisement needs to risk
+    showing a prompt, which exposes it to detection. With this function, the
+    advertisement can silently track just the people who've already granted
+    their location to the newspaper. To defend against this, the UA could
+    indicate when more-sensitive permissions are in use, rather than treating a
+    permission grant as a blank check.
+  </p>
+  <p>
     An adversary could use a <a>permission state</a> as an element in creating a
     "fingerprint" corresponding to an end-user. Although an adversary can
     already determine the state of a permission by actually using the API, that


npdoty:
That text accurately describes the privacy threat, 👍 , but I'm not clear on what mitigations are being suggested (or any that have been implemented in the shipped features).

Ambient notices can be useful for things like location permission in general, but I'm not sure that provides easy detection of these piggybacking uses, especially since the case is a site where the user already explicitly granted a permission.

jyasskin:
I'm not sure there's any way at all to easily detect these piggybacking uses. The ambient notice could identify that the site is using location at an unexpected time, and Chrome's implemented that sort of thing for [media recording and bluetooth](https://cs.chromium.org/chromium/src/chrome/browser/ui/tabs/tab_utils.h?l=29-37&rcl=2e75e0b0f1), but it's certainly not a silver bullet.

Would you rather I drop the "To defend against this" sentence, or replace it with something else?

npdoty:
I think "To defend against this" might overstate the efficacy, but mentioning ambient notice as a mitigation seems helpful. Maybe:

> The UA might provide notice of when permissions are in use on a page which might increase the visibility of abuse.

To be clear, I don't think this resolves #52 at all; the privacy impact of the feature is substantial, even if it's now noted in the spec. 

The ability to subscribe to permission state changes makes the piggybacking even easier to disguise, as the embedded script can access the sensor at the same time as the originating site has received permission.

Issue #182, opened 2018-08-30
A new permission for screen-sharing with getDisplayMedia()


jan-ivar:
Proposing `"display"`. A domain sibling of `"camera"` and `"microphone"`, but simpler, no deviceId stuff.

Also, `"granted"` is forbidden. Its sole purpose will be to persistent blocking.

In https://github.com/w3c/mediacapture-screen-share/pull/73#pullrequestreview-149915990 we imagine this one permissionDescriptor covering both video (screen-sharing) and audio (screen-sharing-related audio, if any).

PR #184, opened 2018-10-11
Add 'display' permission for screen-capture.

jan-ivar:
Fix for https://github.com/w3c/permissions/issues/182.

PTAL @raymeskhoury @marcoscaceres

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/permissions/pull/184.html" title="Last updated on Oct 11, 2018, 9:38 PM GMT (c5ac43a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/184/a58dae5...jan-ivar:c5ac43a.html" title="Last updated on Oct 11, 2018, 9:38 PM GMT (c5ac43a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -623,6 +623,7 @@ spec: webidl
       <a enum-value>"gyroscope"</a>,
       <a enum-value>"magnetometer"</a>,
       <a enum-value>"clipboard"</a>,
+      <a enum-value>"display"</a>,
     };
   </pre>
   <p class="note">
@@ -1003,6 +1004,26 @@ spec: webidl
       permission is the permission associated with the usage of the
       asynchronous methods in the [[clipboard-apis]].
     </p>
+  </section>
+  <section>
+    <h3 id="screen-capture">
+      Screen Capture
+    </h3>
+    <p>
+      The <dfn for="PermissionName" enum-value>"display"</dfn>
+      permission is the permission associated with the usage of
+      [[screen-capture]].
+    </p>
+    <dl>
+      <dt>
+        <a>permission state constraints</a>
+      </dt>
+      <dd>
+        Valid values for this descriptor's <a>permission state</a> are
+        {{"prompt"}} and {{"denied"}}. The user agent MUST NOT ever set this
+        descriptor's <a>permission state</a> to {{"granted"}}.
+      </dd>
+    </dl>
   </section>
 </section>
 <section>


index.bs
```diff
@@ -1004,6 +1005,26 @@ spec: webidl
...
+      permission is the permission associated with the usage of
+      [[screen-capture]].
+    </p>
+    <dl>
+      <dt>
+        <a>permission state constraints</a>
+      </dt>
+      <dd>
+        Valid values for this descriptor's <a>permission state</a> are
+        {{"prompt"}} and {{"denied"}}. The user agent MUST NOT ever set this
```
mounirlamouri: Why should the spec *require* the permission to never be set as granted? Is it a design requirement?

marcoscaceres: Agree. This might be overstepping.

jan-ivar: Yes, it's a design requirement, due to [security risks](https://tools.ietf.org/html/draft-ietf-rtcweb-security-10#section-4.1.1) inherent with screen-capture.

[RTCWEB-SECURITY-ARCH](https://tools.ietf.org/html/draft-ietf-rtcweb-security-arch-15#section-5.2) says: *"Browsers MUST NOT permit permanent screen or application sharing permissions to be installed as a response to a JS request for permissions."*.

jan-ivar: A [blog explainer] (https://blog.mozilla.org/webrtc/share-browser-windows-entire-screen-sites-trust/).

jan-ivar: The [spec prose](https://w3c.github.io/mediacapture-screen-share/#dom-navigator-getdisplaymedia) is already pretty clear, saying: *"The user agent MUST let the end-user choose which display surface to share out of all available choices every time"*.

Adding this language here is necessary to for that spec to call this spec's [prompt the user to choose ](https://w3c.github.io/permissions/#prompt-the-user-to-choose) algorithm, since that algorithm is not guaranteed to prompt the user otherwise.

marcoscaceres: Ok, but this reads like this can never become "granted". That's not the case tho. Maybe this this to be phrased that it can only be "granted" by explicit user consent.

jan-ivar: I think we're fine. I think you're confusing algorithms that *return* `"granted"`, which is fine, with setting the *state*, which is not.

marcoscaceres: Could be. Admittedly, not looked at this in a while.

jan-ivar: The [prompt the user to choose ](https://w3c.github.io/permissions/#prompt-the-user-to-choose) algorithm *"returns either `"denied"` or one of the options."*

jan-ivar:
Thanks @marcoscaceres, can we get a pull of the crank again, so it shows up?

PR #202, opened 2019-11-07
Add periodic-background-sync enum and description.

mugdhalakhani:
I'm not exactly sure where the [web-background-sync] link is coming from, but I think unless there's a spec for Periodic Background Sync, I can't link to that from here. Chicken and egg. PTAL!

@beverloo FYI

***
<a href="https://pr-preview.s3.amazonaws.com/mugdhalakhani/permissions/pull/202.html" title="Last updated on Jun 17, 2021, 4:40 AM UTC (b6355cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/202/deccd89...mugdhalakhani:b6355cc.html" title="Last updated on Jun 17, 2021, 4:40 AM UTC (b6355cc)">Diff</a>


mugdhalakhani:
@mounirlamouri PTAL.

index.bs
```diff
@@ -940,6 +941,16 @@ spec: webidl
...
     </p>
   </section>
+  <section>
+    <h3 id="periodic-background-sync">
+      Periodic Background Sync
+    </h3>
+    <p>
+      The <dfn for="PermissionName" enum-value>"periodic-background-sync"</dfn>
+      permission is the permission associated with the usage of
+      [[web-background-sync]].
```
mounirlamouri: If Periodic Background Sync is an API, you should link to it, not to Background Sync. Otherwise, reading this text, it sounds that there are two permissions meant to control the same API.

mugdhalakhani: Done. Although I'm not able to directly connect to the new spec even after updating the bibio section using:
bikeshed update --biblio

Anyway, I've added a biblio section up top to link to the PBS spec.

index.bs
```diff
@@ -24,6 +24,18 @@ Mailing List Archives: http://lists.w3.org/Archives/Public/public-webappsec/
 
 Markup Shorthands: css no, markdown yes
 </pre>
+
+<pre class="biblio">
+{
```
marcoscaceres: this should live here: https://github.com/WICG/admin/blob/gh-pages/biblio.json

index.bs
```diff
@@ -24,6 +24,18 @@ Mailing List Archives: http://lists.w3.org/Archives/Public/public-webappsec/
...
+  "periodic-background-sync": {
+    "href": "https://wicg.github.io/BackgroundSync/spec/PeriodicBackgroundSync-index.html",
+    "title": "Web Periodic Background Sync",
+    "date": "25 November 2019",
```
marcoscaceres: Probably want to drop the data... it will get out of sync very quickly

index.bs
marcoscaceres: (it will be automatically ingested by SpecRef + BikeShed in 24 hours)

marcoscaceres:
I'm worried about adding this... I don't think we have consensus to make background sync part of the web platform. I think Mozilla it inching towards [not supporting background sync](https://github.com/mozilla/standards-positions/issues/214), and I heard rumors that Apple was going to propose something different (some kind of d/l manager or something).

index.bs
mugdhalakhani: Thanks @marcoscaceres! I'll send a PR to update the WICG/admin code.

marcoscaceres:
Thanks @hober for the link. I think we need to figure out what will become of periodic background sync as a spec before we proceed here. 

There seems to be some renewed interest in background-fetch from multiple browser vendors instead. That's what [I hear](https://github.com/WICG/background-fetch/issues/140#issuecomment-563924166).

mugdhalakhani:
@jakearchibald, @beverloo FYI.

marcoscaceres:
Given that both Mozilla and Apple were pretty firmly against this, I'm going to go ahead and close this.

Issue #175, opened 2018-05-09
Clipboard permissions


martinthomson:
The current clipboard API spec claims that there are two permissions: "clipboard-read" and "clipboard-write".  The current spec includes just one permission: "clipboard".  Which is it?

jyasskin:
@garykac, could you send a PR to make these consistent?

dway123:
I've sent up a PR. As per the Clipboard API spec, there should be 2 separate permissions: "clipboard-read" and "clipboard-write".

marcoscaceres:
As I mentioned in https://github.com/w3c/permissions/pull/209#issuecomment-637899018, I was expecting something like the following, to match other things in the permission spec:

```
enum ClipboardMode { "read", "write" };
dictionary ClipboardPermissionDescriptor : PermissionDescriptor {
   ClipboardMode mode = "read";
};
```

The permission name should remain "clipboard", no?

dway123:
In the Clipboard Specification, it mentions that:

> two separate permissions are defined: "clipboard-read" and "clipboard-write"

From there, these 2 permissions each are further enhanced by a ClipboardPermissionDescriptor, which provides an allowWithoutGesture parameter.

I was thinking it would therefore make sense to have these 2 separate permissions here, with a reference to the Clipboard spec, so that the permission descriptor is defined in the clipboard spec, and doesn't need to be synced between 2 specifications.

Chrome also happens to have implemented these two separate permissions already, and the MDN also documents using these separate permissions.

Changing the Clipboard specification/implementation to have one unified clipboard permission does seem to be more consistent with the existing Permissions spec, but this could break sites already expecting there to be two separate permissions.

dway123:
Would it make sense to add these "clipboard-read" and "clipboard-write" permissions, given my previous message?

marcoscaceres:
Sorry for the delay... I don't think this "-read" and "-write" sets a great precedence. However, I'm sympathetic that Chrome shipped and sites may break.

dway123:
Yeah sorry, I do agree that a single Clipboard permission with the enum you've mentioned would be more consistent with other permissions, and preferable if this were not already reviewed and shipped so long ago.

Thank you for the review and approval. I think this issue can considered resolved and closed now.

marcoscaceres:
> preferable if this were not already reviewed and shipped so long ago.

[...]

PR #209, opened 2020-06-02
Update clipboard permissions as per Clipboard Specification.

dway123:
The [Clipboard API Spec](https://www.w3.org/TR/clipboard-apis/#clipboard-permissions) has had separate `clipboard-read` and `clipboard-write` permissions for a while now, so let's update this spec to mention this. This was mentioned in https://github.com/w3c/permissions/issues/175 as well.

This is my first specification change attempt, so please let me know if there's anything I'm missing. bikeshed spec had no errors when generating this change though.

***
<a href="https://pr-preview.s3.amazonaws.com/dway123/permissions/pull/209.html" title="Last updated on Jun 2, 2020, 11:13 PM UTC (0d78d2f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/209/9675ebf...dway123:0d78d2f.html" title="Last updated on Jun 2, 2020, 11:13 PM UTC (0d78d2f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -622,7 +622,8 @@ spec: webidl
       <a enum-value>"accelerometer"</a>,
       <a enum-value>"gyroscope"</a>,
       <a enum-value>"magnetometer"</a>,
-      <a enum-value>"clipboard"</a>,
+      <a enum-value>"clipboard-read"</a>,
+      <a enum-value>"clipboard-write"</a>,
       <a enum-value>"display-capture"</a>,
       <a enum-value>"nfc"</a>,
     };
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@dway123, thanks, the text looks good... however, this seems to break the convention we have for these kinds of permissions. I would have the value to remain "clipboard" and for a `PermissionDescriptor` to capture to mode. 

Something like: 
```
enum ClipboardMode { "read", "write" }
dictionary ClipboardPermissionDescriptor : PermissionDescriptor {
   ClipboardMode mode = "read";
};
``

dway123:
Thanks for the quick response!

In the Clipboard Specification, it mentions that:
>two separate permissions are defined: "clipboard-read" and "clipboard-write"

From there, these 2 permissions each are further enhanced by a [ClipboardPermissionDescriptor](https://www.w3.org/TR/clipboard-apis/#dictdef-clipboardpermissiondescriptor), which provides an `allowWithoutGesture` parameter.

I was thinking it would therefore make sense to have these 2 separate permissions here, with a reference to the Clipboard spec, so that the permission descriptor is defined in the clipboard spec, and doesn't need to be synced between 2 specifications. 

Chrome also happens to have implemented these two separate permissions already, and the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read) also documents using these separate permissions.

Changing the Clipboard specification/implementation to have one unified clipboard permission does seem to be more consistent with the existing Permissions spec, but this could break sites already expecting there to be two separate permissions.

PR #216, opened 2020-07-20
Noexport the "Set Permission" dfn to fix a Bikeshed warning.

jyasskin:
We can `export`, explicitly `noexport`, or remove the `<dfn>`, and I'm not certain which is best. WebDriver itself gets away without making an explicit choice because it has a local use of every endpoint at https://w3c.github.io/webdriver/#endpoints. (and maybe because Respec doesn't have this warning?) I don't see a reason another specification would depend on this, so `noexport` should be fine, but because it's a public API used by tests, `export` seems semantically right, so I went with `export`.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/permissions/pull/216.html" title="Last updated on Jul 23, 2020, 10:34 PM UTC (3db5fbd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/216/e38762b...jyasskin:3db5fbd.html" title="Last updated on Jul 23, 2020, 10:34 PM UTC (3db5fbd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1090,7 +1090,7 @@ spec: webidl
         </tr>
       </tbody>
     </table>
-    <p>The <dfn export>Set Permission</dfn> <a>extension command</a> simulates user
+    <p>The <dfn noexport>Set Permission</dfn> <a>extension command</a> simulates user
     modification of a {{PermissionDescriptor}}'s <a>permission state</a>.</p>
     <p>The <a>remote end steps</a> are:</p>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
> Respec doesn't have this warning?

Good point... we have a bug to add that.

jyasskin:
Fine with me; switched to `noexport`.

PR #247, opened 2021-06-22
Add 'screen-wake-lock' permission

### LINKED ISSUE #192: Add "screen-wake-lock" permission type
https://github.com/w3c/wake-lock/issues/184

marcoscaceres:
closes #192 

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/29452)

Implementation commitment:

 * [ ] WebKit (link to issue)
 * ✅ Blink 
 * [ ] Gecko (link to issue)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/247.html" title="Last updated on Jul 1, 2021, 6:11 AM UTC (03722d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/247/86e21ca...03722d4.html" title="Last updated on Jul 1, 2021, 6:11 AM UTC (03722d4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -590,6 +590,7 @@ spec:webidl; type:interface; text:Promise
       "notifications",
       "persistent-storage",
       "push",
+      "screen-wake-lock",
       "speaker-selection",
     };
   </pre>
@@ -773,6 +774,16 @@ spec:webidl; type:interface; text:Promise
         </p>
       </dd>
     </dl>
+  </section>
+  <section>
+    <h3 id="screen-wake-lock">
+      Screen wake lock
+    </h3>
+    <p>
+      The <dfn for="PermissionName" enum-value>"screen-wake-lock"</dfn>
+      permission is the permission associated with the usage of the
+      [[screen-wake-lock]] API. It is a <a>boolean feature</a>.
+    </p>
   </section>
   <section>
     <h3 id="media-devices">


marcoscaceres:
@rakuco @kenchris @tomayac, although the `.query({name: "screen-wake-lock"})` works, I noticed that Chrome doesn't actually have a permission prompt for wake lock (on desktop at least) - and it's not listed in the fancy features page for a site. Additionally, it always returns "granted", and provides no way for the user to "deny" the feature? Does it prompt on  on mobile? Will it prompt at all? 

If no, then we might not need this. We can sit on this until we figure that out.

rakuco:
cc @reillyeon

While the current Chrome implementation does go through the permissions system, it also happens to always grant all screen wake lock requests like you noticed. The reasoning back then when we were implementing the code was that existing libraries like NoSleep.js that would keep the screen on with an embedded 1x1 video would not trigger any permission prompts, so we'd try to do something similar for Wake Lock.

I _think_ the embedded video trick does not cause the screen to stay on anymore (I think I saw some reference to it in a bug report months ago, but I can double-check if necessary), but there's been no push in Chrome to change the current wake lock + permissions integration.

With that said, I thought it'd make sense to merge this PR since the Chrome implementation _is_ wired to the Permissions API code?

reillyeon:
This is one of this situations in which we want to reserve the option as implementers to add a permission check later without changing the API in a breaking way. As @rakuco said, Chrome models this as a permission but sets the default value to "allowed" and doesn't have UI for revoking it.

marcoscaceres:
Thanks for clarifying, @reillyeon. I can see the logic in that, particularly as we figure out if we need a permission... however, Chrome might be stuck with it if we decide we don't need a permission, but I guess that's ok. 

@martinthomson, @annevk, Mozilla indicated that the Screen Wake Lock API is "[worth prototyping](https://mozilla.github.io/standards-positions/#screen-wake-lock)", but I understand that it's not a priority at the moment. 

Would it be too advantageous to add this to spec? I could add it and mark it "at risk" waiting on further implementer feedback.

annevk:
@reillyeon is that true for third parties as well? It seems this would be useful for Permissions Policy, at least (with the default being "self").

martinthomson:
I think that the idea we might want controls later, as @reillyeon says, is a good general principle to apply.  If every browser implements the feature AND doesn't use the capability in any way to deny access then we can talk about removing the cruft.  For now, however, it probably makes sense to leave it in.

There is no substantial harm if `.query("blah")` always returns true, for any value of "blah". Identifiers are cheap.

I've also got this vague idea that we might want to have some sort of allowance in the API for heuristic-based grants, particularly those that are granted based on user interaction (which might result in a time-limited grant).

This might be one of those where a browser might want to engage some sort of notification in their UI that the feature is enabled (like fullscreen). In that case, to ensure that the UI gets a chance to be seen, the action might only be permitted in response to user interaction.

(Maybe that's not in scope for this repo, but I thought it worth flagging.)

Anne's point about permissions policy seems like a better reason to keep the permission if you don't like mine :)

rakuco:
> @reillyeon is that true for third parties as well? It seems this would be useful for Permissions Policy, at least (with the default being "self").

The Permissions Policy and Permissions API checks are performed separately, with the former happening first. We follow the spec and default to "self", and `WakeLock.request()` will throw a `NotAllowedError` if the permissions policy check fails.

reillyeon:
> I think that the idea we might want controls later, as @reillyeon says, is a good general principle to apply. If every browser implements the feature AND doesn't use the capability in any way to deny access then we can talk about removing the cruft. For now, however, it probably makes sense to leave it in.
> 
> There is no substantial harm if `.query("blah")` always returns true, for any value of "blah". Identifiers are cheap.
> 
> I've also got this vague idea that we might want to have some sort of allowance in the API for heuristic-based grants, particularly those that are granted based on user interaction (which might result in a time-limited grant).
> 
> This might be one of those where a browser might want to engage some sort of notification in their UI that the feature is enabled (like fullscreen). In that case, to ensure that the UI gets a chance to be seen, the action might only be permitted in response to user interaction.
> 
> (Maybe that's not in scope for this repo, but I thought it worth flagging.)

The idea I've been thinking about is that if the UA displayed a notification that the site is keeping the screen on that could include a button which denied the permission, so we need some spec text describing what happens when the permission is denied. 

> Anne's point about permissions policy seems like a better reason to keep the permission if you don't like mine :)

Mirroring @rakuco's comments, I think that despite the similarity in names Permissions and Permissions Policy use an entirely separate set of identifiers. The specification already has a [Policy Control](https://w3c.github.io/screen-wake-lock/#policy-control) section which completely defines the integration with Permissions Policy.

annevk:
That's correct, but 1) names shouldn't be different between Permissions and Permissions Policy 2) there's a longer term goal of merging these lists and specifications.

rakuco:
> 1) names shouldn't be different between Permissions and Permissions Policy

Just to be clear, this is the case for the Screen Wake Lock API: "screen-wake-lock" is the name of the [powerful feature](https://w3c.github.io/screen-wake-lock/#the-screen-wake-lock-powerful-feature) as well as the [policy-controlled feature](https://w3c.github.io/screen-wake-lock/#policy-control).

marcoscaceres:
@rakuco could you quickly review the associated test at https://github.com/w3c/permissions/issues/192?

rakuco:
Done!

marcoscaceres:
I'm going ahead and merging this as it's implemented in one engine, and there doesn't appear to be any objection to adding it. 

We can go back and revise the decision later if the screen-wake-lock feature itself doesn't get wide adoption.

PR #249, opened 2021-06-29
Handle not fully active documents when querying Permissions API

### LINKED ISSUE #162: Handle non-fully-active documents
It's possible to invoke methods on objects in windows with no longer fully-active documents, because DOM. Generally, these are edge-cases with few if any use-cases other than exploits.

While individual specs are required to think about such cases, very [few do](https://fullscreen.spec.whatwg.org/#dom-element-requestfullscreen).

This presents a unique problem for APIs exposing powerful features behind permissions, because we tie permissions to realms and origins. Specifically, we have to be careful about things like showing the right requesting origin in permission prompts, something that's only guaranteed to make sense from documents that are still [fully active](https://html.spec.whatwg.org/multipage/browsers.html#fully-active).

As an extra safeguard, would it make sense to add prose to the [request permission to use](https://w3c.github.io/permissions/#request-permission-to-use) algorithm to return `"denied"` if the responsible document is no longer *fully active*?

This would expressly forbid powerful features in these edge cases. I can't think of a case where doing so would be a problem.

@jyasskin Thoughts?

marcoscaceres:
closes #162 

Implementation commitment:

 * [ ] WebKit (link to issue)
 * [ ] [Blink]( https://bugs.chromium.org/p/chromium/issues/detail?id=1238709)
 * [ ] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1718598) + [patch](https://phabricator.services.mozilla.com/D118932)

CC @johannhof

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/249.html" title="Last updated on Aug 20, 2021, 6:30 AM UTC (613b424)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/249/da86e19...613b424.html" title="Last updated on Aug 20, 2021, 6:30 AM UTC (613b424)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -197,7 +197,7 @@ spec:webidl; type:interface; text:Promise
         A |descriptor|'s <dfn export local-lt="state">permission state</dfn> for
         an optional <a>environment settings object</a> |settings| is
         the result of the following algorithm, which returns one of
-        {{PermissionState/"granted"}}, {{PermissionState/"prompt"}}, or {{PermissionState/"denied"}}:
+        the {{PermissionState}} enum values or failure:
       </p>
       <ol class="algorithm">
         <li>
@@ -213,6 +213,9 @@ spec:webidl; type:interface; text:Promise
           <code>|descriptor|.{{PermissionDescriptor/name}}</code> and |settings| has an
           <a>associated `Document`</a> named <var>document</var>, run the following step:
           <ol class="algorithm">
+            <li>
+              If |document| is not [=Document/fully active=], return failure.
+            </li>
             <li>
               If <var>document</var> is not <a>allowed to use</a> the feature
               identified by <code>|descriptor|.{{PermissionDescriptor/name}}</code>
@@ -509,7 +512,6 @@ spec:webidl; type:interface; text:Promise
       Promise&lt;PermissionStatus&gt; query(object permissionDesc);
     };
   </pre>
-<<<<<<< HEAD
   <section>
     <h3 id="query-method">`query()` method</h3>
     <p>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Filed [gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1718598) and sent [patch](https://phabricator.services.mozilla.com/D118932).

johannhof:
This seems fine by me by intent though @annevk or so should probably review the spec language :)

marcoscaceres:
@rakina @reillyeon, here is another "not fully active" change 😇 ... I'm wondering if we could coordinate on it? 

I have a patch ready to land in Gecko, but need support from a second implementer before landing it both here and in Gecko.

rakina:
Thanks for doing this! I'm not really familiar with Permissions API, but I'm a bit concerned about not updating the document when it becomes fully active again in this case (Unlike the Geolocation updates in [the other PR](https://github.com/w3c/geolocation-api/pull/90) which seems a bit more OK to drop/leave the document at a stale status), so I've posted in the original issue: https://github.com/w3c/permissions/issues/162#issuecomment-879590670

rakina:
Thanks! This looks reasonable, and I've filed a Chrome bug for this: https://bugs.chromium.org/p/chromium/issues/detail?id=1238709

marcoscaceres:
Thanks @rakina. I'll put together a WTP for this in the meantime.

marcoscaceres:
Moved to https://github.com/w3c/permissions/pull/365

PR #264, opened 2021-08-02
Editorial: Relationship to the Permissions Policy specification

### LINKED ISSUE #176: Delineate relationship between Permissions and Feature Policy
Over in wicg.discourse.io, [ian_clelland writes in response](https://discourse.wicg.io/t/relationship-of-permissions-feature-policy-origin-policy/2772/3?u=equalsjeffh-pp) to [a query regarding the relationship between Permissions and Feature Policy](https://discourse.wicg.io/t/relationship-of-permissions-feature-policy-origin-policy/2772?u=equalsjeffh-pp):

> Fundamentally, [Permissions](https://w3c.github.io/permissions/) is about the user, while [Feature Policy](https://wicg.github.io/feature-policy/) is about the web developer / site owner.

> The Permissions API provides a method for pages to request permission from the user for certain APIs. If the page is allowed to request permission, then the user can choose to grant / deny.

> Feature policy lets developers control which pages/frames are allowed to request permission. By default, scripts running in top-level pages (not inside of <iframes>) have the ability to request permissions from the user, while cross-origin pages in frames are not allowed to request permissions.

I made [the query](https://discourse.wicg.io/t/relationship-of-permissions-feature-policy-origin-policy/2772?u=equalsjeffh-pp) because the above, to me at least, is not clear in the Permissions spec.

marcoscaceres:
closes #176 

@clelland, would you mind taking a look? 

Hopefully this captures the delineation between the two specs correctly. I wonder if we should then add something over on the Permissions Policy spec to complement this?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -561,6 +561,53 @@ spec:webidl; type:interface; text:Promise
     </li>
   </ol>
 </section>
+<section class="non-normative">
+  <h2 id="relationship-to-permissions-policy">
+    Relationship to the Permissions Policy specification
+  </h2>
+  <p>
+    Although technically this specification and the Permissions Policy
+    specification ([[Permissions-Policy]]) deal with "permissions", each
+    specification serves a distinct purpose in the platform. Nevertheless, the
+    two specifications do explicity overlap.
+  </p>
+  <p>
+    On the one hand, this specification exclusively concerns itself with
+    [=powerful features=] whose access is managed through a user-agent mediated
+    permissions UI (i.e., permissions where the user gives express consent
+    before that feature can be used, and where the user retains the ability to
+    deny that permission at any time for any reason). These powerful features
+    are explicitly identified by this specification's {{PermissionName}} enum.
+  </p>
+  <p>
+    On the other hand, the Permissions Policy specification allows developers to
+    selectively enable and disable [=powerful features=] through a
+    "[=permissions policy=]" (be it a HTTP header or a the <{iframe/allow}>
+    attribute). The APIs and features in scope for the Permissions Policy
+    specification go beyond those identified in this specification's
+    {{PermissionName}} enum (e.g., "sync-xhr" and "gamepad"). In that sense, the
+    Permissions Policy subsumes this specification in that Permissions Policy
+    governs wether a feature is available at all, independently of this
+    specification.
+  </p>
+  <p>
+    A powerful feature that has been disabled by the Permissions Policy
+    specification always has its permission state reflected as "denied" by this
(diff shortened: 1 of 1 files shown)


clelland:
@marcoscaceres, I must have missed this in my inbox (overagressive filters, I suppose) -- thanks for writing it up, though! It does a good job of capturing the interplay between the two specifications.

Would it make sense to have parallel text in the Permissions Policy specification as well?

equalsJeffH:
Apologies for missing this, I was on leave across Aug 2021.

WRT:
> Would it make sense to have parallel text in the Permissions Policy specification as well?

I think it would make sense  have parallel text in the Permissions Policy specification.  Note that there is [issue #166](https://github.com/w3c/webappsec-permissions-policy/issues/166) "Delineate relationship between Feature Policy and Permissions" wrt the Permissions Policy specification ;-)

WRT [the text this PR added](https://www.w3.org/TR/permissions/#relationship-to-permissions-policy) to the Permissions spec, it seems to be _only implied_ that where a given feature is _both_ a powerful feature, and is _also_ subject to permissions policy, that the feature's "policy-controlled feature token" (as defined in the feature's specification), and in the Permissions spec's [`PermissionName`](https://www.w3.org/TR/permissions/#dom-permissionname) enum (aka "Powerful features registry"), _can_ be the same, but (I am guessing) do not _have_ to be the same.  Though, I am guessing that the intent is for them to be the same (I cannot quickly find an example where they are not the same). In any case, whatever the requirement is, it ought to be stated explicitly.

E.g., the strings "[camera](https://www.w3.org/TR/permissions/#dom-permissionname-camera)" and "[microphone](https://www.w3.org/TR/permissions/#dom-permissionname-microphone)" are both defined as [powerful feature names](https://www.w3.org/TR/permissions/#dfn-name) and are _also_ declared as [policy-controlled feature tokens (aka "strings")](https://www.w3.org/TR/mediacapture-streams/#permissions-policy-integration).  

Additionally, I find this statement in ([the text this PR added](https://www.w3.org/TR/permissions/#relationship-to-permissions-policy)) confusing:
> The APIs and features in scope for the Permissions Policy specification go beyond those identified in this specification's PermissionName enum (e.g., "sync-xhr" and "gamepad").

...because:  neither "sync-xhr" nor "gamepad" are permission names (i.e., enum values within [`PermissionName`](https://www.w3.org/TR/permissions/#dom-permissionname)) -- so it is not obviously clear what using those policy-controlled feature tokens (aka "strings") as examples is intended to mean.  Clarification would be good.

PR #268, opened 2021-08-30
Fix link to GitHub repository

tidoust:
Links to ED and GitHub repository are currently broken in the spec because of this.

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,7 +33,7 @@
         }],
         github: {
           branch: "main",
-          repoURL: "user/permissions",
+          repoURL: "w3c/permissions",
         },
         mdn: true,
         // See https://respec.org/docs/#xref for usage.


miketaylr:
I filed https://github.com/w3c/permissions/issues/270 to track this (before I saw this PR 😄 )

miketaylr:
And, this was fixed in https://github.com/w3c/permissions/commit/996cf38c61230f80403ced3c0afd39772ced33c8. So it seems like the spec isn't updating correctly.

tidoust:
> So it seems like the spec isn't updating correctly.

Link looks good to me in both https://w3c.github.io/permissions/ and https://www.w3.org/TR/permissions/ so possibly a caching issue? Anyway, thanks for the fix!

miketaylr:
> Link looks good to me in both https://w3c.github.io/permissions/ and https://www.w3.org/TR/permissions/ so possibly a caching issue?

Yeah, think it was a caching issue on my end. Both working fine now. 🥳

PR #284, opened 2021-08-31
Editorial: rename specification

marcoscaceres:
Renaming the spec, making it more clear what it's actually about. 

Also, this spec is commonly referred to as the "Permissions API" (even thought out the spec).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/284.html" title="Last updated on Aug 31, 2021, 5:06 AM UTC (64fe4bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/284/a8d8549...64fe4bc.html" title="Last updated on Aug 31, 2021, 5:06 AM UTC (64fe4bc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>
-      Permissions
+      Permissions API
     </title>
     <script src="https://www.w3.org/Tools/respec/respec-w3c" async class="remove"></script>
     <script class="remove">
@@ -54,9 +54,10 @@
     </script>
   </head>
   <body>
+    <h2 id="subtitle">Interacting with Permissions for Powerful Features</h2>
     <section id="abstract">
       <p>
-        The <cite>Permissions</cite> specification defines common infrastructure that other
+        The <cite>Permissions API</cite> specification defines common infrastructure that other
         specifications can use to interact with browser permissions that allow or deny access to
         powerful features on the web platform. For developers, the specification defines an API to
         query the permission state of a powerful feature, or be notified if a permission for a


miketaylr (approved):
LGTM

jyasskin:
The spec is "actually" about both the web-facing API and the permissions infrastructure that other specifications can use, so naming it after just the API seems misleading. Did this rename involve any of the working group that owns this spec?

miketaylr:
(Apologies if I merged this too quickly.)

marcoscaceres:
No, but I’ll let the group know the rationale for changing it together with an update on the list of recent changes. 

The tl;dr: was to prioritize developers over spec authors, while providing spec authors with the needed infra. 

The spec is colloquially known as the “Permissions API”, so this just aligns with that. 

For example: 
https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API 

https://developers.google.com/web/updates/2015/04/permissions-api-for-the-web

And so on…

jyasskin:
The 2015 WebFundamentals post predates the 2016 change in #97 that added a full set of definitions for other specs to call. The MDN article discusses the API part of this spec (`permissions.query()`) and doesn't name the specification at all. So they're not really evidence that there's a colloquial name for this specification as a whole.

We don't prioritize the web developer interface in naming specs like [Storage](https://storage.spec.whatwg.org/#api) or [Fetch](https://fetch.spec.whatwg.org/#fetch-api), so it's not obvious that we should do it in this one, and doing naming this spec after just its web-facing API incurs the risk that other spec authors will try to reinvent the infrastructure this spec provides. So I'm still a vote for naming it just "Permissions".

miketaylr:
Worse option: rename to "Permissions Infrastructure & API"
Worst option: split out Infra and API into 2 specs.

marcoscaceres:
I don't feel too strongly about it. I'll just switch it back and save us a bunch of bike shedding.

marcoscaceres:
done :)

PR #286, opened 2021-08-31
Rename PermissionName enum to PowerfulFeatureName

marcoscaceres:
The `PermissionName` is a misnomer - it actually identifies "powerful features" for which a permission status can be queries. 

To bring some sanity, I'm renaming `PermissionName` enum to `PowerfulFeatureName`.

 * [x] Modified Web platform tests - change is not observable. 

Implementation commitment:

 * [ ] WebKit (link to issue)
 * [ ] Blink (link to issue)
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1737750)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/286.html" title="Last updated on Oct 26, 2021, 6:55 AM UTC (ff6bc88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/286/50c7149...ff6bc88.html" title="Last updated on Oct 26, 2021, 6:55 AM UTC (ff6bc88)">Diff</a>


marcoscaceres:
Can go with FeatureName - it’s less 💪. Will update.

marcoscaceres:
I might notify various browser vendors of this change before merging, and send patches where appropriate.

jyasskin:
This change will make it more important to unify the list of features between this spec and https://github.com/w3c/webappsec-permissions-policy/blob/main/features.md. It was easier to get away with separate lists when this spec's enum was named something specific to permissions, but if it's named very generally, it needs to actually cover all the "features".

marcoscaceres:
> This change will make it more important to unify the list of features between this spec and https://github.com/w3c/webappsec-permissions-policy/blob/main/features.md. It was easier to get away with separate lists when this spec's enum was named something specific to permissions, but if it's named very generally, it needs to actually cover all the "features".

Right, which is why I originally renamed it to "PowerfulFeatureName". Our spec only deals with APIs that "prompt" (i.e., "powerful features"), not with all the Permission Policy "features". There is only overlap in a small number of cases (e.g., geolocation, notification, push).

I might change it back to the longer name, to make it more clear.

equalsJeffH:
Please see also my belated [comment](https://github.com/w3c/permissions/pull/264#issuecomment-984178181) on PR #264 (relevant portion copied here for convenience):

WRT [the text this PR added](https://www.w3.org/TR/permissions/#relationship-to-permissions-policy) to the Permissions spec, it seems to be _only implied_ that where a given feature is _both_ a powerful feature, and is _also_ subject to permissions policy, that the feature's "policy-controlled feature token" (as defined in the feature's specification), and in the Permissions spec's [`PermissionName`](https://www.w3.org/TR/permissions/#dom-permissionname) enum (aka "Powerful features registry"), _can_ be the same, but (I am guessing) do not _have_ to be the same.  Though, I am guessing that the intent is for them to be the same (I cannot quickly find an example where they are not the same). In any case, whatever the requirement is, it ought to be stated explicitly.

E.g., the strings "[camera](https://www.w3.org/TR/permissions/#dom-permissionname-camera)" and "[microphone](https://www.w3.org/TR/permissions/#dom-permissionname-microphone)" are both defined as [powerful feature names](https://www.w3.org/TR/permissions/#dfn-name) and are _also_ declared as [policy-controlled feature tokens (aka "strings")](https://www.w3.org/TR/mediacapture-streams/#permissions-policy-integration).  

Additionally, I find this statement in ([the text this PR added](https://www.w3.org/TR/permissions/#relationship-to-permissions-policy)) confusing:
> The APIs and features in scope for the Permissions Policy specification go beyond those identified in this specification's PermissionName enum (e.g., "sync-xhr" and "gamepad").

...because:  neither "sync-xhr" nor "gamepad" are permission names (i.e., enum values within [`PermissionName`](https://www.w3.org/TR/permissions/#dom-permissionname)) -- so it is not obviously clear what using those policy-controlled feature tokens (aka "strings") as examples is intended to mean.  Clarification would be good.

marcoscaceres:
Superseded by #353

PR #287, opened 2021-08-31
Define permission lifetimes

### LINKED ISSUE #233: Permission lifetimes
It would be nice if the spec gave guidance on permission lifetimes. See discussion in https://github.com/w3c/geolocation-api/issues/69

marcoscaceres:
closes #233 

Exports [=permission/lifetimes=].

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/287.html" title="Last updated on Nov 18, 2021, 8:01 AM UTC (69a5c32)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/287/c3c5605...69a5c32.html" title="Last updated on Nov 18, 2021, 8:01 AM UTC (69a5c32)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -198,6 +198,11 @@
           </dd>
         </dl>
         <p>
+          A user agent SHOULD provide a means for the user to view, update, and
+          reset the [=permission=] [=permission/state=] of [=powerful features=]
+          associated with a realm or origin.
+        </p>
+        <p>
           To ascertain <dfn class="export">new information about the user's intent</dfn>, a user
           agent MAY collect information about a user's intentions. This information can come from
           explicit user action, aggregate behavior of both the relevant user and other users, or
@@ -212,6 +217,44 @@
             implicit signals.
           </p>
         </aside>
+        <section>
+          <h4>Lifetimes</h4>
+          <p>
+            Every [=permission=] has a <dfn class="export" data-dfn-for="permission">lifetime</dfn>,
+            which is the duration for which a particular permission remains [=permission/granted=].
+            Depending on the [=powerful feature=], the lifetime of a permission is negotiated between the
+            user and the [=user agent=], usually via some permission UI or policy.
+          </p>
+          <p>
+            Specifications that identify themselves as <a>powerful feature</a> SHOULD suggest
+            a [=permission=] [=permission/lifetime=] that is best suited for the particular feature
+            - with a strong emphasis on user privacy.
+          </p>
+          <aside class="Note" title="Determining the lifetime of a permission">
+            <p>
+              For particularly privacy-sensitive [=features=], such as [[[GETUSERMEDIA]]],
+              which can provide access
+              to a user's camera and microphone, user agents are known to expire a permission
+              [=permission/grant=] as soon as a browser tab is closed or navigated.
+              For other features, like the [[[Geolocation]]], user agents are known to offer a choice of
+              only granting the permission for the session, or for one day. Others, like the [[[Notifications]]] and [[[Push]]] APIs,
+              remember a user's decision indefinitely or until the user manually denies the permission.
+            </p>
+            <p>
+              Finding the right balance for the lifetime of a permission requires a lot of
+              thought and experimentation, and often evolves over a long period of time (often years!).
+              Implementers are encouraged to work with their UX security teams to find the right balance
+              between ease of access to a [=powerful feature=] (i.e., reducing the number of permission prompts),
+              respecting a user's privacy, and making users aware when a web application is making use
+              of a particular powerful feature (e.g., via some visual or auditory UI indicator).
+            </p>
+            <p>
+              If you are unsure about what [=permission/lifetime=] to suggest for a [=powerful feature=],
+              please contact the <a href="https://www.w3.org/Privacy/IG/">Privacy Interest Group</a>
+              for guidance.
+            </p>
+          </aside>
+        </section>
       </section>
       <section>
         <h2>
@@ -390,7 +433,8 @@
             A <dfn class="export">default powerful feature</dfn> is a <a>powerful feature</a> with
             all of the above types and algorithms defaulted.
           </p>
-        </section>
+          </section>
+          </p>
         <section>
           <h3>
             Aspects
--- a/index.html
+++ b/index.html
@@ -404,9 +404,9 @@
               <p>
                 Specifications that identify themselves as a [=powerful feature=] SHOULD suggest a
                 [=permission=] [=permission/lifetime=] that is best suited for the particular
-                feature - with a strong emphasis on user privacy. Some guidance on determining the
-                lifetime of a permission is noted below. If no [=permission/lifetime=] is
-                specified, the user agent provides one.
+                feature. Some guidance on determining the lifetime of a permission is noted below,
+                with a strong emphasis on user privacy. If no [=permission/lifetime=] is specified,
+                the user agent provides one.
               </p>
               <p>
                 When the permission [=permission/lifetime=] expires for an origin, and if there are
--- a/index.html
+++ b/index.html
@@ -411,13 +411,17 @@
                 the user agent provides one.
               </p>
               <p>
-                When the permission [=permission/lifetime=] expires for a permission grant to an origin, and if there are
-                [=browsing contexts=] present pertaining to the [=permission=]'s associated origin,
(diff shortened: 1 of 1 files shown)


tomayac:
Woohoo, thanks for specifying this :-) That's it, that's the comment.

index.html
```diff
@@ -209,6 +214,44 @@ <h3>
...
+              to a user's camera and microphone, user agents are known to expire a permission
+              [=permission/grant=] as soon as a browser tab is closed or navigated.
+              For other features, like the [[[Geolocation]]], user agents are known to offer a choice of
+              only granting the permission for the session, or for one day. Others, like the [[[Notifications]]] and [[[push-api]]] APIs,
+              remember a user's decision indefinitely or until the user manually denies the permission.
+            </p>
+            <p>
+              Finding the right balance for the lifetime of a permission requires a lot of
+              thought and experimentation, and often evolves over a long period of time (often years!).
+              Implementers are encouraged to work with their UX security teams to find the right balance
```
miketaylr, line 243: (are UX security teams a thing? or does this mean to say "UX and security"? maybe i just don't know the cool teams 😭)

miketaylr (approved):
LGTM, just a couple of editorial notes.

marcoscaceres:
Thanks @tomayac! It's a start 🙏 Any further feedback you can provide is always welcome! You always provide great feedback.

index.html
marcoscaceres: They are definitely a team. We had one at Mozilla at the Toronto and Vancouver offices and they were awesome. I guess on the Google side, it would be Adrienne Porter Felt and co. (maybe, not sure if they are still doing this kind of work). Happy to rephrase this tho.

index.html
```diff
@@ -209,6 +214,44 @@ <h3>
...
+          <p>
+            Every [=permission=] has a <dfn class="export" data-dfn-for="permission">lifetime</dfn>,
+            which is the duration for which a particular permission remains [=permission/granted=].
+            Depending on the [=powerful feature=], the lifetime of a permission is negotiated between the
+            user and the [=user agent=], usually via some permission UI or policy.
+          </p>
+          <p>
+            Specifications that identify themselves as <a>powerful feature</a> SHOULD suggest
+            a [=permission=] [=permission/lifetime=] that is best suited for the particular feature
+            - with a strong emphasis on user privacy.
```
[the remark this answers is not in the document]
marcoscaceres: Yeah, I'll rejig it.

index.html
miketaylr: Oh yeah, neat. Found their page. No need to rephrase, I think.

index.html
```diff
@@ -209,6 +214,44 @@ <h3>
             implicit signals.
           </p>
         </aside>
+        <section>
+          <h4>Lifetimes</h4>
+          <p>
+            Every [=permission=] has a <dfn class="export" data-dfn-for="permission">lifetime</dfn>,
```
jyasskin, line 220: This needs to get integrated into the algorithms in order to have an effect. You could say that after |lifetime| time since the permission state became "granted", the UA has new information about the user's intent, but that doesn't really seem intuitive. Or you could integrate the lifetime into the definition of "permission state".

jyasskin (changes requested):

index.html
```diff
@@ -382,7 +425,7 @@ <h3>
             A <dfn class="export">default powerful feature</dfn> is a <a>powerful feature</a> with
             all of the above types and algorithms defaulted.
           </p>
-        </section>
+          </section>
```
marcoscaceres, line 428:
```suggestion
        </section>
```

index.html
```diff
@@ -385,6 +390,60 @@ <h3>
...
+                feature. Some guidance on determining the lifetime of a permission is noted below,
+                with a strong emphasis on user privacy. If no [=permission/lifetime=] is specified,
+                the user agent provides one.
+              </p>
+              <p>
+                When the permission [=permission/lifetime=] expires for an origin, and if there are
+                [=browsing contexts=] present pertaining to the [=permission=]'s associated origin,
+                the user agent MUST run the [=powerful feature/permission revocation algorithm=].
+                Alternatively, if there is no [=browsing contexts=] present, the user agent MUST
+                revoke a permission for the origin by setting it back to its default [=permission
```
jyasskin, line 416: "Setting" a permission state isn't really defined in this spec, in order to give UAs lots of freedom of how to infer what the user wants. I think it's clear enough what you mean here, but it could help to define "setting a descriptor's permission state to X" as something like "as if the UA had returned X when reading the descriptor's permission state for some group of settings objects".

index.html
```diff
@@ -385,6 +390,60 @@ <h3>
...
+                  [[[Geolocation]]], user agents are known to offer a choice of only granting the
+                  permission for the session, or for one day. Others, like the [[[Notifications]]]
+                  and [[[push-api]]] APIs, remember a user's decision indefinitely or until the
+                  user manually denies the permission. Note that permission
+                  [=permission/lifetimes=] can vary significantly between user agents.
+                </p>
+                <p>
+                  Finding the right balance for the lifetime of a permission requires a lot of
+                  thought and experimentation, and often evolves over a long period of time (often
+                  years!). Implementers are encouraged to work with their UX security teams to find
```
jyasskin, line 434: Less colloquial maybe:

```suggestion
                  thought and experimentation, and often evolves over a period of years.
                  Implementers are encouraged to work with their UX security teams to find
```

jyasskin (commented):
Thanks, I think this works for integrating the lifetime into the algorithms.

index.html
```diff
@@ -385,6 +390,60 @@ <h3>
...
+              <p>
+                Specifications that identify themselves as a [=powerful feature=] SHOULD suggest a
+                [=permission=] [=permission/lifetime=] that is best suited for the particular
+                feature. Some guidance on determining the lifetime of a permission is noted below,
+                with a strong emphasis on user privacy. If no [=permission/lifetime=] is specified,
+                the user agent provides one.
+              </p>
+              <p>
+                When the permission [=permission/lifetime=] expires for an origin, and if there are
+                [=browsing contexts=] present pertaining to the [=permission=]'s associated origin,
```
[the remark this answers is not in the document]
marcoscaceres: No... I was hesitant to use that word. I'm open to suggestions as to how to make that association (I'll give that some further thought too).

index.html
```diff
@@ -385,6 +390,60 @@ <h3>
...
+                the user agent MUST run the [=powerful feature/permission revocation algorithm=].
```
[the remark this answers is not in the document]
marcoscaceres: Yeah, something like that could work... will add it. 

BTW: I think we need to relook at the "permission revocation algorithm"... it seems to call it self recursively.

index.html
marcoscaceres: Good suggestion.

index.html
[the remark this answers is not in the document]
miketaylr: I think we're trying to say, if you're the top-level browsing context for a given origin and have any ancestor browsing contexts, revoke  permission for all browsing contexts? I guess https://html.spec.whatwg.org/#nested-browsing-contexts gives you most of what you need.

miketaylr (commented):
Looks pretty good to me, pending @jyasskin's comments.

index.html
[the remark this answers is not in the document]
marcoscaceres: It should be possible to treat it like a map, in that we can just grab all any/all browsing contexts associated with an origin (not really worrying if they are nested or not.)

index.html
marcoscaceres: > "Setting" a permission state isn't really defined in this spec, in order to give UAs lots of freedom of how to infer what the user wants. 

I think that's fine, but conceptually there is still some datastore somewhere that gets updated (that is, a permission is bound to an origin or a realm or whatever). When the update happens, it gets asynchronously propagated. 

> I think it's clear enough what you mean here, but it could help to define "setting a descriptor's permission state to X" as something like "as if the UA had returned X when reading the descriptor's permission state for some group of settings objects".

Upon reflection, I think we should do this as a followup, because it gets into the problem above about conceptualizing how permissions are actually stored (or how we will pretend they are stored for the purpose of the specification).

marcoscaceres:
I’m going ahead and merging this as it gives us a good foundation to go off (and it’s blocking geo).

PR #351, opened 2022-02-04
Editorial: clean up Relationship to Permissions Policy

marcoscaceres:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/351.html" title="Last updated on Feb 9, 2022, 5:28 AM UTC (99e7253)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/351/29646e0...99e7253.html" title="Last updated on Feb 9, 2022, 5:28 AM UTC (99e7253)">Diff</a>

PR #350, opened 2022-02-04
Editorial: cleanup examples

marcoscaceres:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/350.html" title="Last updated on Feb 4, 2022, 3:57 PM UTC (4e02d67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/350/29646e0...4e02d67.html" title="Last updated on Feb 4, 2022, 3:57 PM UTC (4e02d67)">Diff</a>

PR #353, opened 2022-02-04
Switch PermissionName to DOMString

### LINKED ISSUE #263: Move permissions out to each spec
The Permissions registry seems a roundabout way to maintain things.... I think we definitely keep the various parts other specs need to define (plus an example of what other specs need to do), keep the enum, but then move all permission descriptions to their respective specs. 

That allows this spec to "gate-keep" the main enum for the API, but allows other specs to define their own permissions. 

TODO:
* [x] #296
* [x] #307
* [x]  "background-fetch", (https://github.com/w3c/permissions/issues/295)
* [x] #308
* [x] #317
* [x] #323
* [x] #322
* [x]  "geolocation" (https://github.com/w3c/geolocation-api/pull/105/)
* [x] #321
* [x] #326
* [x] #294
* [x] #328
* [x] #291
* [x] #329
* [x] #293
* [x] #319
* [x] #320
* [x] #330

### LINKED ISSUE #314: Proposals: Drop "Powerful features registry" and PermissionName enum
I'd like to propose we drop the `PermissionName` enum entirely and switch it to a `required DOMString` instead. 

The rationale for the change is that the Permission Spec is acting as a "gate-keeper" for other specs that may never get broadly supported. 

For example, Gecko only supports the following 4:
https://github.com/mozilla/gecko-dev/blob/43d18f9b44c9967029684aa66d0ad70dee3e89e5/dom/webidl/Permissions.webidl#L10-L16

WebKit has the following as a stub (doesn't actually reflect what is supported):
https://github.com/WebKit/WebKit/blob/cb1203bad562777ee3c25b4eb49ef16477385d1b/Source/WebCore/Modules/permissions/PermissionName.idl#L28-L43

What I propose we do instead is allow each spec to define its power-feature name (e.g., "geolocation"), which is then passed to the algorithms or API as appropriate. We are are already doing in practice.

This change would be backwards compatible, in that implementers could retain their current `PermissionName` enums for implementation purposes. For `.query()`, we would just make `PermissionDescriptor.name` behave like enum. 

It would also mean we no longer need to enum values "at risk" in our spec. 

Any objections?

### LINKED ISSUE #319: Ensure "screen-wake-lock" permission is integrated in Screen Wake Lock API (or just remove it?)

marcoscaceres:
Depends on #350, #351, and the actual Powerful Features Registry.

Closes #319
Closes #314 
Closes #263

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/353.html" title="Last updated on Feb 14, 2022, 6:15 AM UTC (571acf3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/353/acd1296...571acf3.html" title="Last updated on Feb 14, 2022, 6:15 AM UTC (571acf3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -50,6 +50,13 @@
         mdn: true,
         // See https://respec.org/docs/#xref for usage.
         xref: "web-platform",
+        localBiblio: {
+           "powerful-feature-registry": {
+             title: "Registry of Powerful Features",
+             href: "https://w3c.github.io/powerful-features-registry/",
+             status: "unofficial",
+           }
+         }
       };
     </script>
   </head>
--- a/index.html
+++ b/index.html
@@ -50,13 +50,6 @@
         mdn: true,
         // See https://respec.org/docs/#xref for usage.
         xref: "web-platform",
-        localBiblio: {
-          "powerful-feature-registry": {
-            title: "Registry of Powerful Features",
-            href: "https://w3c.github.io/powerful-features-registry/",
-            status: "unofficial",
-          }
-        }
       };
     </script>
   </head>
@@ -75,10 +68,6 @@
     <section id="sotd">
       <p>
         This is a work in progress.
-      </p>
-      <p>
-        Some features in this specification are supported by only one user agent, and as such, are
-        marked as <dfn>at risk</dfn>.
       </p>
     </section>
     <section class="informative">
@@ -577,9 +566,9 @@
         <aside class="issue" id="issue-current-entry-incumbent-or-relevant" data-number="278">
         </aside>
         <p>
-          As a shorthand, a {{PermissionName}} |name|'s <a>permission state</a> is the
-          <a>permission state</a> of a {{PermissionDescriptor}} with its
-          {{PermissionDescriptor/name}} member set to |name|.
+          As a shorthand, a {{DOMString}} |name|'s <a>permission state</a> is the <a>permission
+          state</a> of a {{PermissionDescriptor}} with its {{PermissionDescriptor/name}} member set
+          to |name|.
         </p>
       </section>
       <section>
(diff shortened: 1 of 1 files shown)


miketaylr:
I guess we're going to want to write some patches to Chromium, WebKit, and Gecko as well.

index.html
```diff
@@ -701,6 +701,18 @@ <h4 id="query-method">
...
           <ol class="algorithm">
+            <li>If |permissionDesc|'s {{PermissionDescriptor/name}} member is not supported, return
+            [=a promise rejected with=] a {{TypeError}}.
+              <aside class="note" title="Why is this not an enum?">
+                <p>
+                  This is deliberately design to work the same a WebIDL [=enumeration=] (`enum`)
+                  and implementers are encouraged to use their own custom `enum` here. The reason
+                  this is not an enum in the specification vary greatly in the powerful features
+                  they support. Using a {{DOMString}} to identify a powerful feature gives
+                  implementers the freedom to pick and choose the power features they wish support.
```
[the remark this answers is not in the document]
miketaylr: (tried to clean up a few typos)

marcoscaceres:
> I guess we're going to want to write some patches for Chromium, WebKit, and Gecko as well.

Nah, the point of this change is that they can just use an enum... so nothing need to change in browsers. The change is fully backwards compatible.

miketaylr:
Even better.

index.html
```diff
@@ -726,6 +721,19 @@ <h4 id="query-method">
...
+            <li>If |permissionDesc|'s {{PermissionDescriptor/name}} member is not supported, return
+            [=a promise rejected with=] a {{TypeError}}.
+              <aside class="note" title="Why is this not an enum?">
+                <p>
+                  This is deliberately design to work the same a WebIDL [=enumeration=] (`enum`)
+                  and implementers are encouraged to use their own custom `enum` here. The reason
+                  this is not an enum in the specification vary greatly in the powerful features
+                  they support. Using a {{DOMString}} to identify a powerful feature gives
+                  implementers the freedom to pick and choose which of the power features from the
+                  [[[powerful-feature-registry]]] they wish to support.
```
miketaylr, line 733: This links out to https://w3c.github.io/poweful-features-registry/ (note the typo in powerful) - presumably we fix that elsewhere.

miketaylr (approved):
I think this is good to land now.

index.html
marcoscaceres: Fixed elsewhere (main) and rebased... now made it so once we merge this, it should link without needing the localBiblio.

PR #363, opened 2022-02-15
Editorial: Add linking-text for "specifies a powerful feature"

miketaylr:
Adding this so the link works as expected in https://github.com/w3c/powerful-features-registry/pull/5


miketaylr:
(self-merging while @marcoscaceres is asleep)

PR #365, opened 2022-02-21
Handle non-fully-active documents

### LINKED ISSUE #162: Handle non-fully-active documents

marcoscaceres:
closes #162 

 * [ ] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/32921)

Implementation commitment:

 * [x] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=236952) 
 * [ ] [Blink]( https://bugs.chromium.org/p/chromium/issues/detail?id=1238709)
 * [x] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1718598) + [patch](https://phabricator.services.mozilla.com/D118932)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/pull/365.html" title="Last updated on Feb 24, 2022, 7:30 PM UTC (3bc46a1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/365/4c4eebc...3bc46a1.html" title="Last updated on Feb 24, 2022, 7:30 PM UTC (3bc46a1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -708,6 +708,14 @@
             <var>permissionDesc</var>:
           </p>
           <ol class="algorithm">
+            <li>If [=this=] [=relevant global object=] is a {{Window}} object, then:
+              <ol>
+                <li>If the [=current settings object's=] <a>associated `Document`</a> is not
+                [=Document/fully active=], return [=a promise rejected with=] a
+                {{"InvalidStateError"}} {{DOMException}}.
+                </li>
+              </ol>
+            </li>
             <li>If |permissionDesc|'s {{PermissionDescriptor/name}} member is not supported, return
             [=a promise rejected with=] a {{TypeError}}.
               <aside class="note" title="Why is this not an enum?">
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Put up two tests and a patch for WebKit too. Trying to land Firefox patch in parallel.

kenchris (approved):

index.html
```diff
@@ -837,6 +845,16 @@ <h4>
             steps</dfn>:
           </p>
           <ol class="algorithm">
+            <li>If [=this=] [=relevant global object=] is a {{Window}} object, then:
```
miketaylr, line 848:
```suggestion
            <li>If [=this=]'s [=relevant global object=] is a {{Window}} object, then:
```

index.html
```diff
@@ -837,6 +845,16 @@ <h4>
...
+              <ol>
+                <li>Let |document| this |status|'s [=relevant global object=]'s [=associated
```
miketaylr, line 850:
```suggestion
                <li>Let |document| be |status|'s [=relevant global object=]'s [=associated
```

miketaylr (approved):

PR #376, opened 2022-04-08
Allow "prompt the user to choose" apply to one or more options.

### LINKED ISSUE #375: "Prompt the user to choose" should let caller decide how many results are allowed.
https://w3c.github.io/permissions/#prompt-the-user-to-choose currently insists on picking exactly one of the available options. 

The call @inexorabletash is adding in https://github.com/WICG/local-font-access/pull/76 would like to let the user pick many options, so we should change the wording here to allow that.

https://html.spec.whatwg.org/multipage/input.html#show-the-picker,-if-applicable (for `<input type=file>`), which was a model for the picker design in this spec, also allows users to pick multiple options, so if we ever want to retrofit that to call into the permissions spec, that'll also require this change.

miketaylr:
closes #375

A PR for discussion. PTAL @marcoscaceres

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 23, 2022, 6:26 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fmiketaylr%2Fpermissions%2Fd0c11a15df5aa76e285cafd1925f2d3c9480b527%2Findex.html%3FisPreview%3Dtrue)

```
📡 HTTP Error 404: http://labs.w3.org/spec-generator/uploads/OUQ0MZ
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/permissions%23376.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -662,7 +662,7 @@
           {{PermissionState/"denied"}} and abort these steps.
           </li>
           <li>If |descriptor|'s <a>permission state</a> is {{PermissionState/"granted"}}, the user
-          agent may return one or more of |options| chosen by the user and abort these steps. If the user agent
+          agent may return one (or more if |allowMultiple| is true) of |options| chosen by the user and abort these steps. If the user agent
           returns without prompting, then subsequent <a data-lt="prompt the user to choose">prompts
           for the user to choose</a> from the same set of options with the same |descriptor| must
           return the same option(s), unless the user agent receives <a>new information about the
--- a/index.html
+++ b/index.html
@@ -662,11 +662,11 @@
           {{PermissionState/"denied"}} and abort these steps.
           </li>
           <li>If |descriptor|'s <a>permission state</a> is {{PermissionState/"granted"}}, the user
-          agent may return one (or more if |allowMultiple| is true) of |options| chosen by the user and abort these steps. If the user agent
-          returns without prompting, then subsequent <a data-lt="prompt the user to choose">prompts
-          for the user to choose</a> from the same set of options with the same |descriptor| must
-          return the same option(s), unless the user agent receives <a>new information about the
-          user's intent</a>.
+          agent may return one (or more if |allowMultiple| is true) of |options| chosen by the user
+          and abort these steps. If the user agent returns without prompting, then subsequent
+          <a data-lt="prompt the user to choose">prompts for the user to choose</a> from the same
+          set of options with the same |descriptor| must return the same option(s), unless the user
+          agent receives <a>new information about the user's intent</a>.
           </li>
           <li>Ask the user to choose one or more |options| or deny permission, and wait for them to
           choose:
--- a/index.html
+++ b/index.html
@@ -652,26 +652,26 @@
         </h3>
         <p>
           To <dfn data-lt="prompt the user to choose|prompting the user to choose" class=
-          "export">prompt the user to choose</dfn> one of several |options| associated with a
-          |descriptor|, the user agent must perform the following steps. This algorithm returns
-          either {{PermissionState/"denied"}} or one of the options.
+          "export">prompt the user to choose</dfn> one or more of several |options| associated with
+          a |descriptor|, the user agent must perform the following steps. This algorithm returns
+          either {{PermissionState/"denied"}} or the selected option(s).
         </p>
         <ol class="algorithm">
           <li>If |descriptor|'s <a>permission state</a> is {{PermissionState/"denied"}}, return
           {{PermissionState/"denied"}} and abort these steps.
           </li>
           <li>If |descriptor|'s <a>permission state</a> is {{PermissionState/"granted"}}, the user
-          agent may return one of |options| and abort these steps. If the user agent returns
-          without prompting, then subsequent <a data-lt="prompt the user to choose">prompts for the
-          user to choose</a> from the same set of options with the same |descriptor| must return
-          the same option, unless the user agent receives <a>new information about the user's
-          intent</a>.
+          agent may return one or more of |options| and abort these steps. If the user agent
+          returns without prompting, then subsequent <a data-lt="prompt the user to choose">prompts
+          for the user to choose</a> from the same set of options with the same |descriptor| must
+          return the same option(s), unless the user agent receives <a>new information about the
+          user's intent</a>.
           </li>
-          <li>Ask the user to choose one of the options or deny permission, and wait for them to
-          choose. If the calling algorithm specified extra information to include in the prompt,
-          include it.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -652,26 +652,26 @@ <h3>
...
           </li>
-          <li>Ask the user to choose one of the options or deny permission, and wait for them to
-          choose. If the calling algorithm specified extra information to include in the prompt,
-          include it.
+          <li>Ask the user to choose one or more of the options or deny permission, and wait for
+          them to choose. If the calling algorithm specified extra information to include in the
+          prompt, include it.
           </li>
-          <li>If the user chooses an option, return it; otherwise return
+          <li>If the user choose one or more options, return them; otherwise return
```
jyasskin, line 674: "chooses" or "chose" to agree with "user".

index.html
```diff
@@ -652,26 +652,26 @@ <h3>
         </h3>
         <p>
           To <dfn data-lt="prompt the user to choose|prompting the user to choose" class=
-          "export">prompt the user to choose</dfn> one of several |options| associated with a
-          |descriptor|, the user agent must perform the following steps. This algorithm returns
-          either {{PermissionState/"denied"}} or one of the options.
+          "export">prompt the user to choose</dfn> one or more of several |options| associated with
+          a |descriptor|, the user agent must perform the following steps. This algorithm returns
```
marcoscaceres, line 656:
```suggestion
          a |descriptor|, and optionally a [=non-negative integer=] greater than 1 |restrictions|, which restricts how many |options| a user can select, the user agent must perform the following steps. This algorithm returns
```

index.html
```diff
@@ -652,26 +652,26 @@ <h3>
...
+          either {{PermissionState/"denied"}} or the selected option(s).
         </p>
         <ol class="algorithm">
           <li>If |descriptor|'s <a>permission state</a> is {{PermissionState/"denied"}}, return
```
marcoscaceres, line 660: Then we probably want: 

```suggestion
          <li>Assert: If |restrictions| was passed, |restrictions| is less than or equal to the [=list/size=] of |options|.</li> 
          <li>If |descriptor|'s <a>permission state</a> is {{PermissionState/"denied"}}, return
```

index.html
```diff
@@ -652,26 +652,26 @@ <h3>
         </h3>
         <p>
           To <dfn data-lt="prompt the user to choose|prompting the user to choose" class=
-          "export">prompt the user to choose</dfn> one of several |options| associated with a
-          |descriptor|, the user agent must perform the following steps. This algorithm returns
-          either {{PermissionState/"denied"}} or one of the options.
+          "export">prompt the user to choose</dfn> one or more of several |options| associated with
```
marcoscaceres, line 655:
```suggestion
          "export">prompt the user to choose</dfn> one or more of several [=list/items=] from a [=list=] of |options| associated with
```

index.html
```diff
@@ -652,26 +652,26 @@ <h3>
...
+          agent may return one or more of |options| and abort these steps. If the user agent
+          returns without prompting, then subsequent <a data-lt="prompt the user to choose">prompts
+          for the user to choose</a> from the same set of options with the same |descriptor| must
+          return the same option(s), unless the user agent receives <a>new information about the
+          user's intent</a>.
           </li>
-          <li>Ask the user to choose one of the options or deny permission, and wait for them to
-          choose. If the calling algorithm specified extra information to include in the prompt,
-          include it.
+          <li>Ask the user to choose one or more of the options or deny permission, and wait for
```
marcoscaceres, line 670: Then something like: 
```suggestion
          <li>Ask the user to choose one or more of the options or deny permission. If |restrictions| was passed, then restrict how many |options| to that number. Wait for
```

miketaylr:
(thanks both, will get back to this ASAP)

index.html
jyasskin: FWIW, I'm not sure any callers are going to pick options other than "1" or "unlimited". Those are also the options for `<input>` elements with the [`multiple`](https://html.spec.whatwg.org/multipage/input.html#attr-input-multiple) attribute, which maybe reinforces the idea that they're all we should support here.

miketaylr: Select multiple allows for arbitrary option selection though, if you use the modifier keys (alt/cmd, shift, cmd/ctrl-a, etc): https://miketaylr.com/misc/select-multiple.html

jyasskin: Sorry, I meant that `<input>` allows the caller (the HTML author) to restrict the user to picking 1 option (omit `multiple`) or an unlimited number of options (include `multiple`), but not to restrict the user to picking up to 5 options. It'd be plausible to do the same here.

miketaylr: Aha 💡🧠  - yeah, sounds reasonable.

miketaylr:
OK, re-worked it a bit to take a boolean "allowMultiple". It seems sensible to restrict to one choice in the default case, but allow any number of choices in the non-default case. PTAL @marcoscaceres @jyasskin

marcoscaceres (approved):
Seems good. Let's go with it.

index.html
```diff
@@ -652,26 +652,34 @@ <h3>
...
           <li>If |descriptor|'s <a>permission state</a> is {{PermissionState/"denied"}}, return
           {{PermissionState/"denied"}} and abort these steps.
           </li>
           <li>If |descriptor|'s <a>permission state</a> is {{PermissionState/"granted"}}, the user
-          agent may return one of |options| and abort these steps. If the user agent returns
-          without prompting, then subsequent <a data-lt="prompt the user to choose">prompts for the
-          user to choose</a> from the same set of options with the same |descriptor| must return
-          the same option, unless the user agent receives <a>new information about the user's
-          intent</a>.
+          agent may return one or more of |options| chosen by the user and abort these steps. If the user agent
```
jyasskin, line 665:
```suggestion
          agent may return one (or more if |allowMultiple| is true) of |options| chosen by the user and abort these steps. If the user agent
```

To incorporate `allowMultiple` into the behavior.

jyasskin (approved):

miketaylr:
Thanks for the feedback @jyasskin and @marcoscaceres! 🙇‍♂️

PR #401, opened 2022-11-25
Integrate "request permission to use" with user activation

### LINKED ISSUE #194: Consider making `request-permission-to-use` aware of user activation
The [dfn-obtain-permission](https://w3c.github.io/wake-lock/#dfn-obtain-permission) algorithm could probably take an additional bit in some shape or form prescribing whether user activation is required for showing a permission prompt if the permission state is "prompt".

E.g., the Wake Lock spec currently works around this limitation by essentially re-implementing its own version of this algorithm in [obtain-permission](https://w3c.github.io/wake-lock/#dfn-obtain-permission).

marcoscaceres:
Allows "request permission to use" to accept an optional promise.

The promise gets rejected if the page doesn't have transient activation. Also, when the promise is passed, the algorithm now consumes the user activation. If passed, the promise is resolved with the user's choice.

closes #194

### DIFF
--- a/index.html
+++ b/index.html
@@ -650,33 +650,52 @@
         </h3>
         <p>
           To <dfn data-lt="request permission to use|requesting permission to use" class=
-          "export">request permission to use</dfn> a |descriptor:PermissionDescriptor|, the user
-          agent must perform the following steps. This algorithm returns either
-          {{PermissionState/"granted"}} or {{PermissionState/"denied"}}.
+          "export">request permission to use</dfn> a |descriptor:PermissionDescriptor|, with
+          optionally a |promise:Promise|:
         </p>
         <ol class="algorithm">
+          <li>If |promise| was passed:
+            <ol>
+              <li>Let |global:Window| be |promise|'s [=relevant global object=].
+              </li>
+              <li>Let |document:Document| be |global|'s [=associated `Document`=].
+              </li>
+              <li>If |global| does not have [=transient activation=], [=queue a global task=] on
+              the [=user interaction task source=] given |global| to [=reject=] |promise| with a
+              "{{NotAllowedError}}" {{DOMException}} and abort these steps.
+              </li>
+              <li>Otherwise, [=consume user activation=] of |global|.
+              </li>
+            </ol>
+          </li>
           <li>Let <var>current state</var> be the |descriptor|'s <a>permission state</a>.
           </li>
           <li>If <var>current state</var> is not {{PermissionState/"prompt"}}, return <var>current
           state</var> and abort these steps.
           </li>
-          <li>Ask the user for <a>express permission</a> for the calling algorithm to use the
-          <a>powerful feature</a> described by |descriptor|.
-          </li>
-          <li>If the user gives [=express permission=] to use the powerful feature, return
-          {{PermissionState/"granted"}}; otherwise return {{PermissionState/"denied"}}. The user's
-          interaction may provide <a>new information about the user's intent</a> for this [=global
-          object/realm=] and other [=global object/realms=] with the <a>same origin</a>.
+          <li>Ask the user to choose whether they give [=express permission=] for the [=environment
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Expected use would be something like:

1. Let |promise:Promise| be a [=newly created promise=] in |this|'s [=relevant realm=].
1. Return |promise| and in |parallel|:
   1. [=Request permission=] to use "my-api" with |promise|.
   1. [=React=] to [=promise=]:
     * If it's the case that |promise| is fulfilled with value {{PermissionState/"granted"}}, then:
        1.  ... do granted stuff...
     * If it's the case that |promise| is fulfilled with value {{PermissionState/"denied"}}, then:
        1. ... do denied stuff...

marcoscaceres:
I kinda like this, because it works for legacy APIs (e.g. Geolocation) and for new APIs.

annevk:
The expected usage there is wrong. You cannot react to a promise while in parallel. You'd have to queue a task.

Also, I think Notifications is currently a permission-thingie for which no user activation is required in one implementation (Chromium), but maybe that should be considered a bug.

index.html
```diff
@@ -650,33 +650,52 @@ <h3 id="requesting-more-permission">
...
+            <ol>
+              <li>Let |global:Window| be |promise|'s [=relevant global object=].
+              </li>
+              <li>Let |document:Document| be |global|'s [=associated `Document`=].
+              </li>
+              <li>If |global| does not have [=transient activation=], [=queue a global task=] on
+              the [=user interaction task source=] given |global| to [=reject=] |promise| with a
+              "{{NotAllowedError}}" {{DOMException}} and abort these steps.
+              </li>
+              <li>Otherwise, [=consume user activation=] of |global|.
```
annevk, line 667: I'm not sure this really works if "requesting permission to use" is called while in parallel as the transient activation check really has to happen in the main thread.

index.html
```diff
@@ -650,33 +650,52 @@ <h3 id="requesting-more-permission">
...
+          </li>
+          <li>Otherwise, [=queue a global task=] on the [=user interaction task source=] with
+          |promise|'s [=relevant global object=] to [=resolve=] |promise| with |result|.
+          </li>
         </ol>
         <p>
-          As a shorthand, <a>requesting permission to use</a> a {{DOMString}} |name|, is the same
-          as <a>requesting permission to use</a> a {{PermissionDescriptor}} with its
+          As a shorthand, <a>requesting permission to use</a> a {{DOMString}} |name| and |promise|,
+          is the same as <a>requesting permission to use</a> a {{PermissionDescriptor}} with its
```
miketaylr, line 698: `{{PermissionDescriptor/name}} member set to |name| and |promise|`?

index.html
```diff
@@ -650,33 +650,52 @@ <h3 id="requesting-more-permission">
         </h3>
         <p>
           To <dfn data-lt="request permission to use|requesting permission to use" class=
-          "export">request permission to use</dfn> a |descriptor:PermissionDescriptor|, the user
-          agent must perform the following steps. This algorithm returns either
-          {{PermissionState/"granted"}} or {{PermissionState/"denied"}}.
+          "export">request permission to use</dfn> a |descriptor:PermissionDescriptor|, with
+          optionally a |promise:Promise|:
```
jyasskin, line 654: We need to make call sites easy to read, and I don't think "request permission to use |descriptor| with |promise|" explains to readers what the promise is going to mean. Readers shouldn't need to click through in order to read the call site, which means the call site needs to mention [=transient activation=]. I think something like
```suggestion
          "export">request permission to use</dfn> a |descriptor:PermissionDescriptor|, optionally
          with [=transient activation=] consumed through a |promise:Promise|:
```
would be somewhat better, although "consumed through" is still obscure, and you might come up with better ideas.

annevk: In that case you should define a named parameter: https://infra.spec.whatwg.org/#algorithm-params. (It's not clear to me how a caller would invoke the suggestion.)

PR #407, opened 2022-12-12
Export permissions task source

johannhof:
This is used in the Storage Access API:
https://privacycg.github.io/storage-access/#ref-for-dfn-permissions-task-source

***
<a href="https://pr-preview.s3.amazonaws.com/johannhof/permissions/pull/407.html" title="Last updated on Dec 12, 2022, 11:28 AM UTC (28c0715)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/407/9aaca68...johannhof:28c0715.html" title="Last updated on Dec 12, 2022, 11:28 AM UTC (28c0715)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -455,7 +455,7 @@
           Permissions task source
         </h3>
         <p>
-          The <dfn>permissions task source</dfn> is a [=task source=] used to perform
+          The <dfn class="export">permissions task source</dfn> is a [=task source=] used to perform
           permissions-related [=tasks=] in this specification.
         </p>
       </section>


annevk:
I'm not sure this makes sense. Shouldn't there be a single task queued that informs a document about a state (change) and that task is then used to do various things such as resolve promises or fire events?

As in, I would kinda expect the flow to be that other specifications ask something from Permissions with a set of completion steps and Permissions queues a task and runs those completion steps at the end of that task.

marcoscaceres (approved):

marcoscaceres:
Oops, approved before seeing Anne's comment. I was going to suggest something similar, in that it would be good if the permissions spec did more of the work.

marcoscaceres:
I've also just spotted another incorrect usage of the task source in a different spec... it's being used where the "user interaction task source" should be used.

miketaylr:
Looks like we're using it for `userAgentData.getHighEntropyValues` too: https://wicg.github.io/ua-client-hints/#dom-navigatorua-gethighentropyvalues. I'm guessing the intention was that the browser might reject based on implementation policy, i.e., something like "privacy budget", a user-setting or a "tracking protection" mode enabled. So it's kind of permission-y.

annevk:
Note that using the same task source is really only about relative ordering for the tasks going into that task source.

And since there's only ever really a single task as the result of an in parallel algorithm in the case of permissions, relative ordering might not even matter? I'm not as much of an event loop expert as I'd like to be though, so it would be good if that could be verified.

marcoscaceres (approved):

marcoscaceres:
Ok, closing this... let's make sure the integration points in the spec handle this correctly.

PR #416, opened 2023-06-12
replace cdn.w3.org by www.w3.org

deniak:
Simple PR to replace cdn.w3.org by www.w3.org (cdn.w3.org is going down) on the official ED.
[Respec has already been fixed](https://github.com/w3c/respec/pull/4442).


miketaylr (approved):
lgtm, thanks

PR #420, opened 2023-09-14
Inform users when a UA administrator overrides their permission choices.

jyasskin:
This is still somewhat exploratory, to see if this is a good way to explain how UAs should handle cases where their enterprise policies override user choices about permissions. It implements https://w3ctag.github.io/privacy-principles/#device-administrators for this spec, and the idea to try it out here came from the [Standardizing managed user agent behavior session at TPAC 2023](https://www.w3.org/events/meetings/d98c0762-eb0c-4cf6-85eb-fff31dedc9e9/).

@reillyeon, can you cc in whichever Chrome enterprise folks should pay attention? @timcappalli are you the right person to check this for Microsoft's enterprise system?

Implementation commitment:

 * [ ] WebKit (link to issue)
 * [ ] Blink (link to issue)
 * [ ] Gecko (link to issue)

### DIFF
--- a/index.html
+++ b/index.html
@@ -234,7 +234,8 @@
             <dfn class="export" data-local-lt="grant" data-dfn-for="permission">Granted</dfn>:
           </dt>
           <dd>
-            The user, or the user agent on the user's behalf, has given [=express permission=] to
+            The user, or the user agent or its [=administrator=] on the user's behalf,
+            has given [=express permission=] to
             use a [=powerful feature=]. The caller will can use the feature possibly without having
             the [=user agent=] asking the user's permission.
           </dd>
@@ -242,7 +243,8 @@
             <dfn class="export" data-dfn-for="permission">Denied</dfn>:
           </dt>
           <dd>
-            The user, or the user agent on the user's behalf, has denied access to this [=powerful
+            The user, or the user agent or its [=administrator=] on the user's behalf,
+            has denied access to this [=powerful
             feature=]. The caller will can't use the feature.
           </dd>
         </dl>
@@ -550,9 +552,11 @@
             |settings|, returning |previousResult|, and the user agent has not received <a>new
             information about the user's intent</a> since that invocation, return |previousResult|.
             </li>
-            <li>Return the instance of |name|'s [=powerful feature/extra permission data type=]
-            that matches the UA's impression of the user's intent, taking into account any
-            [=powerful feature/extra permission data constraints=] for |name|.
+            <li>Return the instance of |name|'s [=powerful feature/extra permission data type=] that
+            matches the UA's impression of the user's intent, combined with any configuration from
+            the user agent's [=administrator=], taking into account any [=powerful feature/extra
+            permission data constraints=] for |name|. If [=administrator=] configuration affected the
+            result, the [=user agent=] must inform the user.
             </li>
           </ol>
           <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -550,9 +552,11 @@ <h2>
...
             </li>
-            <li>Return the instance of |name|'s [=powerful feature/extra permission data type=]
-            that matches the UA's impression of the user's intent, taking into account any
-            [=powerful feature/extra permission data constraints=] for |name|.
+            <li>Return the instance of |name|'s [=powerful feature/extra permission data type=] that
+            matches the UA's impression of the user's intent, combined with any configuration from
+            the user agent's [=administrator=], taking into account any [=powerful feature/extra
+            permission data constraints=] for |name|. If [=administrator=] configuration affected the
+            result, the [=user agent=] must inform the user.
             </li>
```
jyasskin, line 560: I think this is how to integrate Chrome's [WebUsbAllowDevicesForUrls](https://chromeenterprise.google/policies/#WebUsbAllowDevicesForUrls) policy, which I believe adds to the set of devices that https://wicg.github.io/webusb/#dom-usb-getdevices returns by default. @reillyeon might correct me.

This text says to inform the user if configuration affects the result either by adding or removing things. Is that right, or should we only inform the user if admin configuration adds to the page's abilities?

index.html
```diff
@@ -817,18 +821,31 @@ <h3 id="requesting-more-permission">
...
           </li>
-          <li>Ask the user for <a>express permission</a> for the calling algorithm to use the
-          <a>powerful feature</a> described by |descriptor|.
+          <li>If the [=user agent=] has an [=administrator=] who has configured the [=user agent=]
+          to grant or deny permission for the calling algorithm to use the <a>powerful feature</a>
+          described by |descriptor|:
+          <ol>
+            <li>Set |current state| to {{PermissionState/"granted"}} if the [=administrator=]
+            granted permission; otherwise to {{PermissionState/"denied"}}.</li>
+            <li>Inform the user of the [=administrator=]'s configuration.</li>
```
jyasskin, line 830: Similarly here, should we only inform the user if the admin configuration forced a 'grant' result?

index.html
```diff
@@ -1308,6 +1325,12 @@ <h2 id="privacy-considerations">
         A user agent SHOULD provide a means for the user to review, update, and reset the
         [=permission=] [=permission/state=] of [=powerful features=] associated with an [=origin=].
       </p>
+      <p>
+        Some [=user agents=] support <dfn data-lt="administrator">administrators</dfn> who can
```
jyasskin, line 1329: I expect the `<dfn>` to move somewhere else, perhaps Infra, in the future, but I think the discussion about putting it there will be easier if it's already used in a spec or two first.

index.html
```diff
@@ -817,18 +821,31 @@ <h3 id="requesting-more-permission">
           <li>If <var>current state</var> is not {{PermissionState/"prompt"}}, return <var>current
           state</var> and abort these steps.
           </li>
-          <li>Ask the user for <a>express permission</a> for the calling algorithm to use the
-          <a>powerful feature</a> described by |descriptor|.
+          <li>If the [=user agent=] has an [=administrator=] who has configured the [=user agent=]
```
jyasskin, line 824: This implies that administrators are always using enterprise policy, while https://w3ctag.github.io/privacy-principles/#dfn-administrator implies that a user who owns their own device is that device's administrator. We could make either meaning work, but this meaning seems to produce simpler spec text. If we go with this, we should eventually rewrite the Privacy Principles to match.

index.html
```diff
@@ -1308,6 +1325,12 @@ <h2 id="privacy-considerations">
         A user agent SHOULD provide a means for the user to review, update, and reset the
         [=permission=] [=permission/state=] of [=powerful features=] associated with an [=origin=].
       </p>
+      <p>
+        Some [=user agents=] support <dfn data-lt="administrator">administrators</dfn> who can
+        override user choices about permissions, either to block permissions or to allow them on
+        certain sites. Overriding user choices in this way can be a privacy problem, so this
+        specification requires that the [=user agent=] inform users when it happens.
```
miketaylr, line 1332: Out of curiosity, do any browsers in managed environments expose admin set permissions today? 

On my work laptop, I see the following in settings in Chrome:
Screenshot 2023-09-15 at 3 30 56 PM

And the following in Firefox settings:

Screenshot 2023-09-15 at 3 31 00 PM

index.html
reillyeon: Yes, enterprise policy adds additional entries to the result of `getDevices()`. As implemented it cannot remove entries but I could see that as a possibility.

index.html
jyasskin: Chrome has a couple that are exposed like 
calendar.google.com site info panel showing Notifications, Allowed, and "Managed by your organization" and 
"HID devices with usages from usage page F1D0" allowed to chrome-extension://ckcendljdlmgnhghiaomidhiiclmapok/. This setting is managed by your administrator.

marcoscaceres:
This feels like it should be a Privacy Consideration rather than affecting the algorithms.

index.html
marcoscaceres: This seems to assume the policy is being applied in real time?

jyasskin:
I don't think we should do the whole thing as a COMEFROM from the Privacy Considerations, but we could move some of the acknowledgement of the administrator's existence to, say, https://pr-preview.s3.amazonaws.com/jyasskin/permissions/pull/420.html#dfn-permission. Then the discussion of how to notify the user could live in the Privacy Considerations. Does that make more sense to you @marcoscaceres?

Issue #424, opened 2023-11-06
Umbrella / Meta: Browser Permissions for WebDriver BiDi as an extension module


thiagowfx:
WebDriver BiDi spec umbrella issue: https://github.com/w3c/webdriver-bidi/issues/587

"Interacting with Permissions for Powerful Features"

The [**WebDriver BiDi spec**](https://w3c.github.io/webdriver-bidi/) should support interaction with **browser permissions** as an [extension module](https://w3c.github.io/webdriver-bidi/#extension-modules). These permissions represent a user's choice to allow or deny access to "powerful features" of the platform. For developers, the specification standardizes an API to query the permission state of a powerful feature, and be notified if a permission to use a powerful feature changes state.

The Permissions spec already supports [WebDriver classic](https://www.w3.org/TR/permissions/#set-permission-command). This bug is about making it support [WebDriver BiDi](https://w3c.github.io/webdriver-bidi/) as well.

[For Puppeteer it would be nice to add as the next step](https://github.com/w3c/permissions/pull/425#pullrequestreview-1734377396):

# Tracking

- [ ] MPV / Initial PR: https://github.com/w3c/permissions/pull/425
- [ ] TODO: Add `user context` to `SetPermissionParameters`
- [ ] TODO: Add `origin` to `SetPermissionParameters`

thiagowfx:
cc @OrKoN @mathiasbynens

PR #425, opened 2023-11-06
Introduce Browser Permissions for WebDriver BiDi

thiagowfx:
Based-on: https://github.com/w3c/webdriver-bidi/pull/590
Bug: #424
Bug: https://github.com/w3c/webdriver-bidi/issues/587

Upstream auto-linking blockers:

- [x] https://github.com/w3c/webdriver-bidi/pull/593
- [x] https://github.com/w3c/webdriver-bidi/pull/591
- [x] https://github.com/w3c/webdriver/pull/1766
- [x] https://github.com/w3c/webdriver/pull/1767

Other considerations:

- [x] https://www.w3.org/2023/11/08-webdriver-minutes.html#t04
- ~~<jgraham_> (don't want to extend this topic further, but I always assumed BiDi would move to an event-based model for permissions i.e. you'd get a permissions.Request event with some information)~~:
 
This is out of scope of this initial PR. We intend to only achieve feature parity with WebDriver classic.

- - -

 * [x] ~~Modified~~ Web platform tests _commitment_ (link): https://github.com/web-platform-tests/wpt/issues/43305

Implementation commitment:

 * [ ] WebKit (link to issue)
 * [x] Blink (actually, Chromium BiDi mapper): https://github.com/GoogleChromeLabs/chromium-bidi/issues/1585
 * [x] Gecko (link to issue): https://github.com/mozilla/standards-positions/issues/930

***
<a href="https://pr-preview.s3.amazonaws.com/thiagowfx/permissions/pull/425.html" title="Last updated on Nov 22, 2023, 1:24 PM UTC (0183692)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/425/f647e6d...thiagowfx:0183692.html" title="Last updated on Nov 22, 2023, 1:24 PM UTC (0183692)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1382,7 +1382,6 @@
                     permissions.SetPermissionParameters = {
                       descriptor: permissions.PermissionDescriptor,
                       state: permissions.PermissionState,
-                      ? context: browsingContext.BrowsingContext,
                       ? origin: text,
                     }
                   </pre>
@@ -1408,28 +1407,13 @@
    parameters|.
                 </li>
                 <li>
-                  Let |context id| be the value of the <code>context</code> field of
-   |command parameters|.
-                </li>
-                <li>
-                  If |context id| is not present, let |context| be the [=top-level browsing context=]
-   of |session|. Otherwise, let |context| be the result of [=trying=] to
-   [=get a browsing context=] with |context id|.
-                </li>
-                <li>
                   Let |permission name| be the value of the <code>name</code> field of
    |descriptor| representing {{PermissionDescriptor/name}}.
                 </li>
                 <li>
-                  <!-- TODO: Only convert to an IDL value only once. -->
-                  [=Convert to an IDL value=] (|descriptor|, |state|) of type {{PermissionSetParameters}}.
-   If this conversion throws an exception, return [=error=] with [=error code=] [=invalid argument=].
-                </li>
-                <li>
-                  Let |typedDescriptor| be the object |descriptor| refers to, <a>converted to an IDL
-                    value</a> of |permission name|'s [=powerful
-                    feature/permission descriptor type=]. If this throws an exception,
-                    return [=error=] with [=error code=] [=invalid argument=].
+                  Let |typedDescriptor| be the object |descriptor| refers to, [=converted to an IDL value=] (|descriptor|, |state|) of
+                  {{PermissionSetParameters}} |permission name|'s [=powerful feature/permission descriptor type=].
+                  If this conversion throws an exception, return [=error=] with [=error code=] [=invalid argument=].
                 </li>
                 <li>
                   If |state| is an inappropriate [=permission state=] for any
--- a/index.html
+++ b/index.html
@@ -1277,7 +1277,7 @@
               </p>
             </li>
             <li>Let |typedDescriptor| be the object |rootDesc| refers to, <a>converted to an IDL
-            value</a> of `|rootDesc|.{{PermissionDescriptor/name}}`'s [=powerful
+            value</a> of |rootDesc|.{{PermissionDescriptor/name}}'s [=powerful
             feature/permission descriptor type=]. If this throws an exception, return a [=invalid
             argument=] [=error=].
             </li>
--- a/index.html
+++ b/index.html
@@ -1308,63 +1308,63 @@
         This document defines the following [=extension modules=] for the [[WebDriver-BiDi]] specification.
       </p>
       <section>
-        <h4 id="extension-module-browser">
-          The browser Module
+        <h4 id="extension-module-permissions">
+          The permissions Module
         </h4>
         <p>
-          The [=modules/browser=] module contains commands for
-          managing the remote end browser process.
+          The <dfn>permissions</dfn> module contains commands for
+          managing the remote end browser permissions.
         </p>
         <section>
-          <h5 id="extension-module-browser-definition">
+          <h5 id="extension-module-permissions-definition">
             Definition
           </h5>
           <p>
             [=remote end definition=]
           </p>
           <pre class="cddl remote-cddl">
-            BrowserCommand = (
-              browser.setPermission
+            PermissionsCommand = (
+              permissions.setPermission
             )
           </pre>
         </section>
         <section>
-          <h5 id="extension-module-browser-types">
+          <h5 id="extension-module-permissions-types">
             Types
           </h5>
           <section>
-            <h6 id="extension-module-type-browser-PermissionDescriptor">
-              The browser.PermissionDescriptor Type
+            <h6 id="extension-module-type-permissions-PermissionDescriptor">
+              The permissions.PermissionDescriptor Type
             </h6>
             <pre class="cddl local-cddl">
-              browser.PermissionDescriptor = {
+              permissions.PermissionDescriptor = {
                 type: "permissionDescriptor",
                 name: text,
               }
             </pre>
             <p>
-              The <code>browser.PermissionDescriptor</code> type represents a {{PermissionDescriptor}}.
+              The <code>permissions.PermissionDescriptor</code> type represents a {{PermissionDescriptor}}.
             </p>
           </section>
           <section>
-            <h6 id="extension-module-type-browser-PermissionState">
-              The browser.PermissionState Type
+            <h6 id="extension-module-type-permissions-PermissionState">
+              The permissions.PermissionState Type
             </h6>
             <pre class="cddl local-cddl">
-              browser.PermissionState = "granted" / "denied" / "prompt"
+              permissions.PermissionState = "granted" / "denied" / "prompt"
             </pre>
             <p>
-              The <code>browser.PermissionState</code> type represents a {{PermissionState}}.
+              The <code>permissions.PermissionState</code> type represents a {{PermissionState}}.
             </p>
           </section>
         </section>
         <section>
-          <h5 id="extension-module-browser-commands">
+          <h5 id="extension-module-permissions-commands">
             Commands
           </h5>
           <section>
-            <h6 id="extension-module-command-browser-setPermission">
-              The browser.setPermission Command
+            <h6 id="extension-module-command-permissions-setPermission">
+              The permissions.setPermission Command
             </h6>
             <p>
               The <dfn class="export" data-dfn-for="extension modules">Set Permission</dfn>
@@ -1375,14 +1375,14 @@
               <dt>Command Type</dt>
               <dd>
                 <pre class="cddl remote-cddl">
-                  browser.setPermission = (
-                    method: "browser.setPermission",
-                    params: browser.SetPermissionParameters
+                  permissions.setPermission = (
+                    method: "permissions.setPermission",
+                    params: permissions.SetPermissionParameters
                   )
 
-                  browser.SetPermissionParameters = {
-                    descriptor: browser.PermissionDescriptor,
-                    state: browser.PermissionState,
+                  permissions.SetPermissionParameters = {
+                    descriptor: permissions.PermissionDescriptor,
+                    state: permissions.PermissionState,
                     ? context: browsingContext.BrowsingContext,
                     ? origin: text,
                   }
--- a/index.html
+++ b/index.html
@@ -1371,8 +1371,8 @@
                 The permissions.setPermission Command
               </h6>
               <p>
-                The <dfn class="export" data-dfn-for="extension modules">Set Permission</dfn>
-                [=extension module=] simulates user modification of a {{PermissionDescriptor}}'s
+                The <dfn class="export">Set Permission</dfn>
+                [=command=] simulates user modification of a {{PermissionDescriptor}}'s
                 [=permission state=].
               </p>
               <dl>
@@ -1399,10 +1399,10 @@
              </dl>
              <ol algorithm="remote end steps for permissions.setPermission">
                <p>
-                TODO: Add user context to SetPermissionParameters.
+                TODO: Add user <code>context</code> to <code>SetPermissionParameters</code>.
                </p>
                <p>
-                TODO: Add origin.
+                TODO: Add <code>origin</code>.
                </p>
                <p>
                  The [=remote end steps=] with |session| and |command parameters| are:
--- a/index.html
+++ b/index.html
@@ -1177,14 +1177,14 @@
         permission=] or by some user agent policy.
       </p>
     </section>
-    <section data-cite="webdriver">
+    <section>
       <h2 id="automation">
         Automated testing
       </h2>
       <p>
         For the purposes of user-agent automation and application testing, this document defines
-        the following <a>extension commands</a> for the [[WebDriver]] specification. It is OPTIONAL
-        for a user agent to support <a>extension commands</a> commands.
+        extensions to the [[WebDriver]] and [[WebDriver-BiDi]] specifications.
+        It is OPTIONAL for a user agent to support them.
       </p>
       <pre class='idl'>
         dictionary PermissionSetParameters {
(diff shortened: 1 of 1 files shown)


thiagowfx:
cc @OrKoN @mathiasbynens @jgraham @whimboo (I don't have permissions to change the "assignee" field directly)

index.html
```diff
@@ -1177,119 +1177,305 @@ <h2 id="relationship-to-permissions-policy">
...
+                  </pre>
+                </dd>
+                <dt>Result Type</dt>
+                <dd>
+                  <pre class="cddl">
+                    EmptyResult
+                  </pre>
+                </dd>
+             </dl>
+             <div algorithm="remote end steps for permissions.setPermission">
```
OrKoN, line 1290: would it make sense to extract the common algorithm for actually setting the permissions? we probably do not want to just duplicate the setting steps

thiagowfx: But the variables are different in BiDi and classic. How do we "bind" them in the same algorithm?

thiagowfx:
cc @miketaylr could you PTAL?

index.html
```diff
@@ -1177,119 +1177,305 @@ <h2 id="relationship-to-permissions-policy">
...
-          <a>extension command</a> simulates user modification of a {{PermissionDescriptor}}'s
-          <a>permission state</a>.
+          This document defines the following [=extension modules=] for the [[WebDriver-BiDi]] specification.
         </p>
+        <section>
+          <h4 id="extension-module-browser">
+            The browser Module
+          </h4>
+          <p>
+            The [=modules/browser=] module contains commands for
```
OrKoN, line 1207: I am not sure we can redefine the whole module like this. From the BiDi spec:

https://w3c.github.io/webdriver-bidi/#protocol-modules

```
An implementation may define extension modules. These must have a module name that contains a single colon ":" character. The part before the colon is the prefix; this is typically the same for all extension modules specific to a given implementation and should be unique for a given implementation. Such modules extend the local end definition and remote end definition providing additional groups as choices for the defined commands and events.
```

At least we should discuss if we need to extend that defintion.

index.html
```diff
@@ -1177,119 +1177,305 @@ <h2 id="relationship-to-permissions-policy">
...
           required PermissionDescriptor descriptor;
           required PermissionState state;
         };
       </pre>
-      <section>
-        <h3 id="set-permission-command">
-          Set Permission
+      <section data-cite="webdriver-bidi webdriver">
+        <h3 id="automation-webdriver-bidi">
+          Automated testing with [[WebDriver-BiDi]]
```
miketaylr, line 1197: nit: Why would WebDriver-Bidi come before WebDriver? If I were sorting them lexicographically, I would place WebDriver first. The diff would be a lot easier to review as well.

index.html
```diff
@@ -1177,119 +1177,305 @@ <h2 id="relationship-to-permissions-policy">
...
-                /session/{session id}/permissions
-              </td>
-            </tr>
-          </tbody>
-        </table>
         <p>
-          The <dfn class="export" data-dfn-for="extension commands">Set Permission</dfn>
-          <a>extension command</a> simulates user modification of a {{PermissionDescriptor}}'s
-          <a>permission state</a>.
+          This document defines the following [=extension modules=] for the [[WebDriver-BiDi]] specification.
```
miketaylr, line 1200: It seems like this dfn is not exported.

miketaylr (commented):
Noting that if you were to place WebDriver-Bidi after WebDriver, this would be a bit easier to review.

Also, can you fix the build errors?

index.html
thiagowfx: Done

index.html
thiagowfx: There was a bug in webref: https://github.com/w3c/webref/issues/1071

Which is now fixed, so now the `dfn` is exported.

thiagowfx:
Converting this back to draft until I apply all requested changes (both here and via https://www.w3.org/2023/11/08-webdriver-minutes.html#t04).

index.html
miketaylr: Ah, great. Thanks! 🐞 🐛

index.html
```diff
@@ -1177,119 +1177,306 @@ <h2 id="relationship-to-permissions-policy">
...
+                  <pre class="cddl remote-cddl">
+                    permissions.setPermission = (
+                      method: "permissions.setPermission",
+                      params: permissions.SetPermissionParameters
+                    )
+
+                    permissions.SetPermissionParameters = {
+                      descriptor: permissions.PermissionDescriptor,
+                      state: permissions.PermissionState,
+                      ? context: browsingContext.BrowsingContext,
```
OrKoN, line 1385: todo (as discussed): this would be a user context but we could leave it out at first as it would be a new feature compared with the webdriver classic

thiagowfx: Done. Removed.

index.html
thiagowfx: There's no AI on this end of the spec. The AI is to change the BiDi spec to remove the colon requirement. Therefore, resolving this comment.

index.html
OrKoN: So looks like the bidi steps 6 to 10 and classic steps 5 to 9 are the same and could be extracted into an algorithm

index.html
miketaylr: Is there an issue filed to track removing the colon requirement?

index.html
miketaylr: +1, you can either create a new variable to capture `parameters.state` somewhere before step 4 (in classic), and pass it into an algorithm with a better name than `to 'foo bar' with state`, or just pass in `parameters.state` directly to said algorithm.

index.html
jgraham: There is no "colon requirement". Note that paragraph starts "An _implementation_ may…". This specification is not an implementation, so it doesn't apply.

However the BiDi spec currently doesn't talk about external specifications at all (and presumably couldn't do so in a binding way). However if it did, it would express the preference that third party specifications add additional modules where possible rather than adding to existing commands, events or modules, just so that it's clear what the total definition of each module is.

index.html
```diff
@@ -1177,119 +1177,294 @@ <h2 id="relationship-to-permissions-policy">
...
+            <h5 id="extension-module-permissions-types">
+              Types
+            </h5>
+            <section>
+              <h6 id="extension-module-type-permissions-PermissionDescriptor">
+                The permissions.PermissionDescriptor Type
+              </h6>
+              <pre class="cddl local-cddl">
+                permissions.PermissionDescriptor = {
+                  type: "permissionDescriptor",
```
thiagowfx, line 1340: Is this needed? Can I remove it?

index.html
thiagowfx: Done.

index.html
OrKoN: Can we make it `permissions.PermissionDescriptor = text`?

index.html
```diff
@@ -1177,119 +1177,270 @@ <h2 id="relationship-to-permissions-policy">
...
+                  If |state| is an inappropriate [=permission state=] for any
+   implementation-defined reason, return [=error=] with [=error code=] [=invalid argument=].
+                </li>
+                <li>
+                  Let |typedDescriptor| be the object |descriptor| refers to, [=converted to an IDL value=] (|descriptor|, |state|) of
+                  {{PermissionSetParameters}} |permission name|'s [=powerful feature/permission descriptor type=].
+                  If this conversion throws an exception, return [=error=] with [=error code=] [=invalid argument=].
+                </li>
+                <li>
+                  Run the [=set a permission=] algorithm, passing |typedDescriptor| and |state|.
```
OrKoN, line 1434:
```suggestion
                  [=Set a permission=] with |typedDescriptor| and |state|.
```

?

thiagowfx: Done

index.html
thiagowfx: It would be problematic if we wanted to extend it in the future, e.g. with the fields of https://chromedevtools.github.io/devtools-protocol/tot/Browser/#type-PermissionDescriptor

thiagowfx:
FYI: Most of the CI errors are pre-existing:
image

The only error stemming from this PR is the last one. But it's a deadlock. The CI complains about `<div class="algorithm">`, and it also complains about `<ol>` inside `<ol>`. If you have a third attempt idea, let me know.

index.html
OrKoN: Ack, what did you want to remove? the type? I think it can be removed.

thiagowfx: Yes, just the type. OK, I'll remove it then.

thiagowfx: Done.

index.html
thiagowfx: Should we leave the BiDi spec as is then, w.r.t. this colon statement?

OrKoN: We should only (optionally) add a recommendation about how external specifications should be authored (e.g., defining complete modules). I think we can leave the text about implementation extensions as is. It looks like there is nothing blocking on the WebDriver BiDi spec side.

thiagowfx: I'll resolve this here and add an AI to https://github.com/w3c/webdriver-bidi/issues/587

thiagowfx:
> This looks good to me. For Puppeteer it would be nice to add origins (or browsing contexts?) and the user context as the next step.

Noting this as an AI here: https://github.com/w3c/permissions/issues/424

thiagowfx:
This is ready to go.

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+          <section>
+            <h5 id="extension-module-permissions-commands">
+              Commands
+            </h5>
+            <section>
+              <h6 id="extension-module-command-permissions-setPermission">
+                The permissions.setPermission Command
+              </h6>
+              <p>
+                The <dfn class="export" data-dfn-for="extension modules">Set Permission</dfn>
```
OrKoN, line 1374: I think we should remove the link to the extension module definition. You can link to https://w3c.github.io/webdriver-bidi/#module instead

thiagowfx: Done

miketaylr:
> Most of the CI errors are pre-existing:

Hm, I just sent in a PR last week and didn't have any CI issues (https://github.com/w3c/permissions/pull/426)

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+                result of an invocation of <a>permission state</a> for |descriptor| with the
+                argument |target| made at this moment.
+                </li>
+              </ol>
+            </li>
+            <li>[=list/Append=] |task| to |tasks|.
+            </li>
+          </ol>
+        </li>
+        <li>Wait for all <a>tasks</a> in |tasks| to have executed.
```
miketaylr, line 1222:
```suggestion
        <li>Wait for all <a>tasks</a> in |tasks| to have executed and return.
```

miketaylr: maybe?

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
-          </tbody>
-        </table>
         <p>
-          The <dfn class="export" data-dfn-for="extension commands">Set Permission</dfn>
-          <a>extension command</a> simulates user modification of a {{PermissionDescriptor}}'s
-          <a>permission state</a>.
+          This document defines the following <a>extension commands</a> for the [[WebDriver]] specification.
         </p>
+        <section>
+          <h4 id="set-permission-command">
```
miketaylr, line 1233:
```suggestion
          <h4 id="webdriver-set-permission-command">
```

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+                <li>
+                  Let |typedDescriptor| be the object |descriptor| refers to, [=converted to an IDL value=] (|descriptor|, |state|) of
+                  {{PermissionSetParameters}} |permission name|'s [=powerful feature/permission descriptor type=].
+                  If this conversion throws an exception, return [=error=] with [=error code=] [=invalid argument=].
+                </li>
+                <li>
+                  [=Set a permission=] with |typedDescriptor| and |state|.
+                </li>
+                <li>
+                  Return [=success=] with data null.
```
miketaylr, line 1436:
```suggestion
                  Return [=success=] with data `null`.
```

miketaylr (changes requested):
Looks pretty good, but we do need to fix the markup errors before we can merge.

marcoscaceres:
Remember we still need implementation commitments etc. before we can merge. 

We also need to figure out what we need to add over on the WPT side.

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
       <pre class='idl'>
         dictionary PermissionSetParameters {
           required PermissionDescriptor descriptor;
           required PermissionState state;
         };
       </pre>
-      <section>
-        <h3 id="set-permission-command">
-          Set Permission
+      <p>To <dfn>set a permission</dfn> given <var>descriptor</var>, and <var>state</var>:
```
marcoscaceres, line 1195:
```suggestion
      <p>To <dfn>set a permission</dfn> given |descriptor:PermissionDescriptor|, and |state:PermissionState|:
```

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+      <ol class="algorithm">
+        <li>Let |settings| be the [=current settings object=].
+        </li>
+        <li>Let |targets| be a <a>list</a> containing all <a>environment settings objects</a>
```
marcoscaceres, line 1199:
```suggestion
        <li>Let |targets| be a <a>list</a> containing all [=environment settings objects=]
```

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+        whose [=environment settings object/origin=] is the <a data-lt="same origin">same</a> as
+        the [=environment settings object/origin=] of |settings|.
```
marcoscaceres, line 1201:
```suggestion
        whose [=environment settings object/origin=] is [=same origin=] as
        the [=environment settings object/origin=] of |settings|.
```

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+        </li>
+        <li>Let |tasks| be an empty <a>list</a>.
+        </li>
+        <li>For each <a>environment settings object</a> |target| in |targets|:
```
marcoscaceres, line 1205: Where possible, please use the [= =] syntax... 

```suggestion
        <li>For each [=environment settings object=] |target| in |targets|:
```

index.html
marcoscaceres: Yeah, what's the intent here with waiting?

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+          <ol>
+            <li>
+              <a>Queue a task</a> |task| on the [=permissions task source=] of |target|'s
```
marcoscaceres, line 1208: This is still a little confusing... I was expecting:

Let |task| be ... steps...
[=Queue a task=] |task| ....

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+            Set Permission
+          </h4>
+          <table>
+            <tbody>
+              <tr>
+                <th>
+                  HTTP Method
+                </th>
+                <th>
+                  <a data-lt="extension command URI template">URI Template</a>
```
marcoscaceres, line 1243:
```suggestion
                  [=extension command URI template|URI Template=]
```

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+            type {{PermissionSetParameters}}. If this throws an exception, return an [=invalid
+            argument=] [=error=].
+            </li>
+            <li>Let |rootDesc| be |parameters|.{{PermissionSetParameters/descriptor}}.
+            </li>
+            <li>If |parameters|.{{PermissionSetParameters/state}} is an inappropriate <a>permission
+            state</a> for any implementation-defined reason, return a [=invalid argument=] [=error=].
+              <p class="note">
+                For example, <a>user agents</a> that define the "midi" <a>powerful feature</a> as
+                "always on" may choose to reject a command to set the <a>permission state</a> to
```
marcoscaceres, line 1275:
```suggestion
                "always on" can choose to reject a command to set the <a>permission state</a> to
```

How would they do that though? It can only be set to "granted",   "denied",  or "prompt" ?

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+                {{PermissionState/"denied"}} at this step.
+              </p>
+            </li>
+            <li>Let |typedDescriptor| be the object |rootDesc| refers to, <a>converted to an IDL
+            value</a> of <code>|rootDesc|.{{PermissionDescriptor/name}}</code>'s [=powerful
```
marcoscaceres, line 1280:
```suggestion
            value</a> of |rootDesc|.{{PermissionDescriptor/name}}'s [=powerful
```

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+            <li>Return <a>success</a> with data `null`.
+            </li>
+          </ol>
+          <aside class="example" title="Setting a permission via WebDriver">
+            <p>
+              To [=extension commands/set permission=] for `{name: "midi", sysex: true}` of the
+              [=current settings object=] of the <a>session</a> with ID 23 to "`granted`", the local
+              end would POST to `/session/23/permissions` with the body:
+            </p>
+            <pre class="lang-json">
```
marcoscaceres, line 1296:
```suggestion
            <pre class="json">
```

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+              <h6 id="extension-module-type-permissions-PermissionDescriptor">
+                The permissions.PermissionDescriptor Type
+              </h6>
+              <pre class="cddl local-cddl">
+                permissions.PermissionDescriptor = {
+                  name: text,
+                }
+              </pre>
+              <p>
+                The <code>permissions.PermissionDescriptor</code> type represents a {{PermissionDescriptor}}.
```
marcoscaceres, line 1350:
```suggestion
                The `permissions.PermissionDescriptor` type represents a {{PermissionDescriptor}}.
```

marcoscaceres: Same with the others

marcoscaceres (changes requested):
Couple of suggestions still to fix... but nice to see this coming along.

index.html
thiagowfx: In theory, `<code>` and backticks are synonyms. I chose `<code>` to mirror the pattern of the BiDi spec.

In this case, I'll replace all `<code>` occurrences with backticks. Should I also replace the occurrences that this PR did not touch? (fine by me)

thiagowfx: Done

index.html
thiagowfx: Done

index.html
thiagowfx: Done

index.html
thiagowfx: Note: This is part of pre-existing prose. Updated anyway.

index.html
thiagowfx: This is part of pre-existing prose: https://www.w3.org/TR/permissions/#set-permission-command

Should I make the proposed change?

index.html
thiagowfx: There are 64 occurrences of `<a>`. Better do this in another PR, agreed? I'm afraid it would be too noisy to change them to `[= =]` in this PR.

index.html
thiagowfx: Done

index.html
thiagowfx: Done

index.html
thiagowfx: This is pre-existing prose: https://www.w3.org/TR/permissions/#set-permission-command

It seems to me the intent is to just "await" before returning from the method/function, for syncing purposes

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+                </dd>
+                <dt>Result Type</dt>
+                <dd>
+                  <pre class="cddl">
+                    EmptyResult
+                  </pre>
+                </dd>
+             </dl>
+             <ol algorithm="remote end steps for permissions.setPermission">
+               <p>
```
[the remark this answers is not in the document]
thiagowfx: Changed outer tag to `<div>` to match the BiDi style. Fixed.

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+                TODO: Add user <code>context</code> to <code>SetPermissionParameters</code>.
+               </p>
+               <p>
+                TODO: Add <code>origin</code>.
+               </p>
+               <p>
+                 The [=remote end steps=] with |session| and |command parameters| are:
+               </p>
+               <ol>
```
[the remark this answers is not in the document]
thiagowfx: Changed outer tag to `<div>` to match the BiDi style. Fixed.

index.html
```diff
@@ -1177,119 +1177,269 @@ <h2 id="relationship-to-permissions-policy">
...
+                <dt>Result Type</dt>
+                <dd>
+                  <pre class="cddl">
+                    EmptyResult
+                  </pre>
+                </dd>
+             </dl>
+             <ol algorithm="remote end steps for permissions.setPermission">
+               <p>
+                TODO: Add user <code>context</code> to <code>SetPermissionParameters</code>.
```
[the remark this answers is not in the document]
miketaylr: OK, I think that's fine, but would prefer to not merge bare TODO text in the spec. :)

Could you file issues on the repo instead and link to them here, i.e., https://respec.org/docs/#issue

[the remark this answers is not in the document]
thiagowfx: I am fine to remove the TODOs here. The reason I added them in the first place was the precedent on the BiDi spec itself: https://w3c.github.io/webdriver-bidi/ -- it currently has 16 TODOs.

But of course, if the Permissions spec doesn't follow this convention, we can just keep track of them in the Issues.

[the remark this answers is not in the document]
miketaylr: That would be my preference, thanks. Inline issues are totally fine, and useful. So feel free to add them per the respec link.

thiagowfx:
@jgraham @whimboo Does Mozilla / Firefox / Gecko intend to support the browser permissions API for BiDi?

If yes, then could you please file a bug and edit the first comment to include it?
If not, then I will edit it myself to add a note that there's no implementation intent.

[Extension modules](https://w3c.github.io/webdriver-bidi/#extension-modules) are optional anyway.

thiagowfx:
@gsnedders Does Apple / WebKit intend to support the browser permissions API for BiDi?

If yes, then could you please file a bug and edit the first comment to include it?
If not, then I will edit it myself to add a note that there's no implementation intent.

[Extension modules](https://w3c.github.io/webdriver-bidi/#extension-modules) are optional anyway.

index.html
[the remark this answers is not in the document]
thiagowfx: Done. Moved to a trackling checklist on #424

index.html
thiagowfx: Ah, now I see the point of adding the `and return`, because it's extracted into an algorithm. Done.

index.html
thiagowfx: Done, instead: `webdriver-command-set-permission`

index.html
thiagowfx: Done

index.html
thiagowfx: Done, thanks! Adding the types looks good

index.html
thiagowfx: I changed the occurrence you asked for, but won't change many others so that we don't add a lot of noise in the PR

index.html
thiagowfx: Done

thiagowfx:
I believe all suggestions were addressed. PTAL

miketaylr:
> This branch has conflicts that must be resolved

https://www.youtube.com/watch?v=RlJM8-7IMBM

(let me try to clean this up before reviewing)

thiagowfx:
If the rebase ends up being straightforward, feel free to merge it (or I can do it).

If there are any follow-ups, Alex (orkon) will take over from here. Thanks!

OrKoN:
I have opened a PR with the rebase here https://github.com/w3c/permissions/pull/431 since I have no permissions to @thiagowfx fork. Let's close this PR in favor of the new PR.

miketaylr:
> I have opened a PR with the rebase here #431 since I have no permissions to @thiagowfx fork. Let's close this PR in favor of the new PR.

Sounds good.

PR #444, opened 2024-02-13
Use `object` for PermissionSetParameters.descriptor

### LINKED ISSUE #443: Turn PermissionSetParameters.descriptor into an object
The Set Permission step 3 does: https://w3c.github.io/permissions/#webdriver-command-set-permission

>3. Let rootDesc be parameters.[descriptor](https://w3c.github.io/permissions/#dom-permissionsetparameters-descriptor).

But `parameters` is a JS object here, so it can't just access the IDL property but should do some JS `Get()`, as Streams does in https://streams.spec.whatwg.org/#abstract-opdef-setupcrossrealmtransformreadable.

But an even better way to fix it is just to turn `decriptor` as an object, which would allow `Let rootDesc be parametersDict.descriptor`.

(Step 4 does the same thing for `rootDesc.name` when `rootDesc` is a JS object, maybe it can be "[permission descriptor type](https://w3c.github.io/permissions/#dfn-permission-descriptor-type) per the result of `Get(rootDesc, "name")`"

saschanaz:
And also fixes rootDesc.name use. Failing to convert `descriptor` would throw at a later step but this shouldn't matter as all error types are same.

closes #443

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/permissions/pull/444.html" title="Last updated on Feb 14, 2024, 10:20 PM UTC (f0a61b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/permissions/444/12a5113...saschanaz:f0a61b9.html" title="Last updated on Feb 14, 2024, 10:20 PM UTC (f0a61b9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1225,7 +1225,7 @@
       </p>
       <pre class='idl'>
         dictionary PermissionSetParameters {
-          required PermissionDescriptor descriptor;
+          required object descriptor;
           required PermissionState state;
         };
       </pre>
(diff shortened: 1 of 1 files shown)


annevk (commented):
This looks reasonable, but this is not an editorial change. This fixes several actual issues.

marcoscaceres (approved):

miketaylr (approved):

marcoscaceres:
Yes, this is fine, but we need to potentially update implementations to match the IDL change.

miketaylr:
> Yes, this is fine, but we need to potentially update implementations to match the IDL change.

Looks like it for Blink: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/interfaces/permissions.idl;l=39?q=PermissionSetParameters

marcoscaceres:
Actually, as it's just part of the WebDriver part, I think it's ok to merge. @miketaylr, is that ok with you?

Issue #388, opened 2022-10-25
Allow returning "prompt" rather than "denied"


annevk:
"denied" at times reveals information about prior visits that user agents might not be willing to expose. For certain permissions we might want to categorically outlaw it if all user agents are in agreement.

"denied" should be fine to return when it's the result of Permissions Policy or when lacking a secure context. I don't think that needs changing.

For other cases, user agents should be permitted to return "prompt", probably implementation-defined? But also, for certain permissions all implementations might be in agreement to not return "denied" (except as per above), such as "storage-access". The specification should provide infrastructure for that as well as it'll leave less up to implementations.

marcoscaceres:
Yeah, totally makes sense. We should definitely add this.

johannhof:
@miketaylr mind if I take a stab at this?

johannhof:
Actually, can't storage-access just provide its own [`permission query algorithm`](https://www.w3.org/TR/permissions/#dfn-permission-query-algorithm)? Was there anything beyond that you were hoping to achieve here, @annevk? I suppose we could still add a note that UAs can always run implementation defined steps to hide/change the return value?

annevk:
I didn't know that existed. https://github.com/privacycg/storage-access/issues/60#issuecomment-1278022204 is what I based this issue on. That might be sufficient if `query()` is the only way this state ends up being exposed.

I guess we should consider whether we want to do be doing that more generally, but that's probably best tracked separately as a follow-up.

(I'm a little wary of all these pluggable algorithms though as what actually ends up happening is not exactly clear.)

annevk:
After discussing this a bit further I think there is a difference. The current setup is such that the internal algorithm is not pluggable, but the public API is. This means that specifications that build on top of this will still see "denied" (due to the end user having denied; seeing "denied" for other reasons is not a problem as that state can already be observed via other means). This has some minor fallout:

1. Specifications building on top of Permissions might end up exposing "denied" (do to the end user having denied) through a side channel. Any such case should probably be considered a bug, but it would be impossible if they couldn't observe that state to begin with.
2. Attackers could probably "Spectre read" the state. This is not a very problematic "Spectre read" attack, but it's also avoidable.

[...]

PR #458, opened 2025-01-27
Allow for the query algorithm to return `prompt` or `denied` when document is not `allowed to use`

aselya:
Updated language to allow the query algorithm to return `prompt` or`denied` helps protect the user from exposing their available features and helps prevent retaliation against the user from developers.

closes [#388](https://github.com/w3c/permissions/issues/388)

Implementation commitment:

 * [ ] WebKit (link to issue)
 * [ ] Blink (link to issue)
 * [ ] Gecko (link to issue)

### DIFF
--- a/index.html
+++ b/index.html
@@ -800,7 +800,14 @@
                 [=associated `Document`=].
                 </li>
                 <li>If <var>document</var> is not <a>allowed to use</a> |feature|, return
-                {{PermissionState/"denied"}}.
+                {{PermissionState/"denied"}} or {{PermissionState/"prompt"}}.
+
+                <p class="note">
+                  The {{PermissionState/"prompt"}} may be returned instead of
+                  {{PermissionState/"denied"}} to avoid exposing if the |feature| is
+                  <a>allowed to use</a> to developers. This is done to prevent retaliation against
+                  the user and repeated prompting to the detriment of the user experience.
+              </p>
                 </li>
               </ol>
             </li>


miketaylr:
There is some prior art at https://privacycg.github.io/storage-access/#permissions-integration and https://privacycg.github.io/requestStorageAccessFor/#permissions-integration

Conceptually, WDYT @marcoscaceres, should we pull this into permissions, or just add a note saying powerful features can do this in their own permission query algorithms?

johannhof:
@aselya can you elaborate a bit more on why you think exposing Permissions Policy state (which is "allowed to use") would  lead to retaliation against the user?

I could see an argument for why this technically exposes cross-origin information, but that seems by design, the same way that, say, the sandbox argument is observable by a cross-origin iframe. Also, that doesn't seem like something that should be implementation-defined. :)

index.html
```diff
@@ -800,7 +800,14 @@ <h3 id="reading-current-states">
...
                 <li>If <var>document</var> is not <a>allowed to use</a> |feature|, return
-                {{PermissionState/"denied"}}.
+                {{PermissionState/"denied"}} or {{PermissionState/"prompt"}}.
+
+                <p class="note">
+                  The {{PermissionState/"prompt"}} may be returned instead of
+                  {{PermissionState/"denied"}} to avoid exposing if the |feature| is
+                  <a>allowed to use</a> to developers. This is done to prevent retaliation against
+                  the user and repeated prompting to the detriment of the user experience.
+              </p>
```
marcoscaceres, line 810:
```suggestion
                <p class="note">
                  For privacy reasons, In certain cases, the user agent can return th4 {{PermissionState/"prompt"}} state instead of {{PermissionState/"denied"}}. 
              </p>
```

index.html
```diff
@@ -800,7 +800,14 @@ <h3 id="reading-current-states">
                 [=associated `Document`=].
                 </li>
                 <li>If <var>document</var> is not <a>allowed to use</a> |feature|, return
-                {{PermissionState/"denied"}}.
+                {{PermissionState/"denied"}} or {{PermissionState/"prompt"}}.
```
marcoscaceres, line 803:
```suggestion
                {{PermissionState/"denied"}}.
     </li>
     <li>Optionally, return {{PermissionState/"prompt"}}.
```

marcoscaceres:
Ok, yeah, @johannhof is right (this has nothing to do with "allowed to use")... this needs to happen further down around... we think around step 8, where the permissions store is checked. We reviewed this and added some suggested text for the note by incorporating some of Anne's wording.

aselya:
> @aselya can you elaborate a bit more on why you think exposing Permissions Policy state (which is "allowed to use") would lead to retaliation against the user?
> 
> I could see an argument for why this technically exposes cross-origin information, but that seems by design, the same way that, say, the sandbox argument is observable by a cross-origin iframe. Also, that doesn't seem like something that should be implementation-defined. :)

Apologies for the delay in response, only just saw this.

I made this PR after observing that this spec and the [spec](https://privacycg.github.io/requestStorageAccessFor/#permissions-integration) for requestStorageAccessFor (rSAFor) were not in alignment on the what permission states might be returned from a query. The explanation provided in the rSAFor spec for not revealing the [denied](https://w3c.github.io/permissions/#dfn-denied) permission state seemed reasonable and worth incorporating into the permissions spec to allow for other permissions to utilize in the same manner.

aselya:
@marcoscaceres, moving the text to step 8 makes sense to me.

aselya:
@marcoscaceres @miketaylr, any additional questions preventing this merge?

annevk (commented):
I don't really see how this addresses https://github.com/w3c/permissions/issues/388#issuecomment-1323946527

In particular:

> we keep this issue as a follow-up to make "permission query algorithm" run at a lower-level (or possibly not expose "denied" (do to the end user having denied) at all anymore).

I guess this note is meant to indicate user agents can do whatever, but is that really desirable?

